PR #1, opened 2019-04-23
Move to Docker Hub registry

nmarley:


nmarley:
See also #2

shumkov:
@nmarley We could do everything in #2 ?

nmarley:
We should try and stick to the same PR guidelines as Dash Core:

https://github.com/dashpay/dash/blob/master/CONTRIBUTING.md#pull-request-philosophy

> In general commits should be atomic and diffs should be easy to read. For this reason do not mix any formatting fixes or code moves with actual code changes.

> Patchsets should always be focused. For example, a pull request could add a feature, fix a bug, or refactor code; but not a mixture. Please also avoid super pull requests which attempt to do too much, are overly large, or overly complex as this makes review difficult.

PR #5, opened 2020-03-06
feat: update configurations and docker-compose file

shuplenkov:


README.md
```diff
@@ -1,109 +1,50 @@
...
-$ ./mn-bootstrap.sh regtest logs
-```
+Package contains several configuration presets:
+ - Local - standalone masternode for local development
+ - Evonet - masternode with Evonet configuration
 
-If you want to call dash-cli inside of the dashd container, you can use the `dash-cli.sh` wrapper
+There are two ways to apply a present:
+ 1. Rename corresponding dotenv file (i.e. `.env.local`) to `.env`
+ 2. Add `--env-file` option to `docker-compose` command (requires `docker-compose` v1.25.0+)
```
shumkov: @thephez Does it make sense to specify minimal versions in "Pre-requisites" section?

thephez: Yes, good idea. I overlooked that. :+1:

PR #11, opened 2020-03-19
feat: remove spork private key from Core config

shuplenkov:


configs/evonet/core/dashd.conf
```diff
@@ -42,9 +42,6 @@ printtoconsole=1
 
 addnode=34.220.38.116:20001
 
-sporkaddr=yMtULrhoxd8vRZrsnFobWgRTidtjg2Rnjm
```
shumkov: We need spork address

PR #15, opened 2020-03-21
feat: add testnet preset

strophy:
This splits the `docker-compose.yml` file in two and specifies testnet configs so we can only load evo services when we are on a network with evo support. First step towards being able to use this tool as a dashman replacement.


.env.evonet
```diff
@@ -8,5 +8,7 @@ CORE_MASTERNODE_BLS_PRIV_KEY=
 CORE_EXTERNAL_IP=
 DPNS_CONTRACT_ID=77w8Xqn25HwJhjodrHW133aXhjuTsTv9ozQaYpSHACE3
 DPNS_TOP_LEVEL_IDENTITY=77w8Xqn25HwJhjodrHW133aXhjuTsTv9ozQaYpSHACE3
+RPC_PORT=20002
```
shumkov: why do we changing RPC ports?

strophy: Good point, removed

.env.testnet
```diff
@@ -0,0 +1,9 @@
+COMPOSE_PROJECT_NAME="dash_masternode_testnet"
+PRESET=testnet
+CORE_VERSION=latest
+CORE_MASTERNODE_BLS_PRIV_KEY=
+CORE_EXTERNAL_IP=
+# testnet or mainnet
+SENTINEL_NETWORK=testnet
+RPC_PORT=19998
```
shumkov: here as well

.env.evonet
```diff
@@ -8,5 +8,6 @@ CORE_MASTERNODE_BLS_PRIV_KEY=
 CORE_EXTERNAL_IP=
 DPNS_CONTRACT_ID=77w8Xqn25HwJhjodrHW133aXhjuTsTv9ozQaYpSHACE3
 DPNS_TOP_LEVEL_IDENTITY=77w8Xqn25HwJhjodrHW133aXhjuTsTv9ozQaYpSHACE3
+P2P_PORT=20001
```
shumkov: Let's add a prefix to the var (`CORE_`)

docker-compose.yml
```diff
@@ -5,7 +5,7 @@ services:
     image: dashpay/dashd:${CORE_VERSION:-latest}
     restart: always
     ports:
-      - 20001:20001 # P2P
+      - ${P2P_PORT}:${P2P_PORT} # P2P
```
shumkov: Let's keep internal as is then you can always use the same port to connect in other services.

strophy: OK, so internally we always use 20001 for P2P and 20002 for RPC? Are you sure this will work for mainnet masternodes? I remember there was a check to ensure port 9999 is used at some point, but if it is only during `register_prepare` then we will be specifying the external port there anyway.

shumkov: Hm.. I think you right. It’s better to set right P2P port to Dash Core. Anyways, we are not using P2P port in other services atm.

.env.testnet
```diff
@@ -0,0 +1,8 @@
+COMPOSE_PROJECT_NAME="dash_masternode_testnet"
+PRESET=testnet
+CORE_VERSION=latest
```
shumkov: Not sure is good idea to use latest. Using specific version we can force to update + more reliable coz we control it. In other presets we set 0.15

README.md
```diff
@@ -28,17 +28,18 @@ Package contains Docker Compose file and configuration presets.
...
+ - Testnet - masternode with testnet configuration
 
 There are two ways to apply a present:
  1. Rename corresponding dotenv file (i.e. `.env.local`) to `.env`
  2. Add `--env-file` option to `docker-compose` command
 
 ### Start
 
-In order to run a masternode use Docker Compose:
+In order to run a masternode use Docker Compose (optionally with evo services):
```
shumkov: I would add one more section above about Evo configuration and explained that there are two ways to use Evo setup:
1. Provide `-f docker-compose.yml -f docker-compose.evo.yml` every time you use `docker-compose` command
2. Create symlink `ln -s docker-compose.evo.yml docker-compose.override.yml` (or just copy?). A symlink is better from the perspective of future mn-bootstrap updates when we change evo yml

.env.evonet
```diff
@@ -8,5 +8,6 @@ CORE_MASTERNODE_BLS_PRIV_KEY=
 CORE_EXTERNAL_IP=
 DPNS_CONTRACT_ID=77w8Xqn25HwJhjodrHW133aXhjuTsTv9ozQaYpSHACE3
 DPNS_TOP_LEVEL_IDENTITY=77w8Xqn25HwJhjodrHW133aXhjuTsTv9ozQaYpSHACE3
+CORE_P2P_PORT=20001
 # testnet or mainnet
 SENTINEL_NETWORK=testnet
```
shumkov: actually we could use `COMPOSE_FILE` env in Evo present in order to specify proper config files:
```
COMPOSE_FILE=docker-compose.yml:docker-compose.evo.yml
```

README.md
shumkov: I found a better solution with COMPOSE_FILE so we could remove all this stuff at all.

.env.testnet
strophy: I set this to latest because codablock usually runs binaries from the core `develop` branch in his testnet containers, since this is what is being tested.

docker-compose.yml
```diff
@@ -5,14 +5,13 @@ services:
...
     volumes:
-      - ./configs/${PRESET:?err}/core/dashd.conf:/dashd.conf
-      - ./data/${PRESET:?err}/core:/data
+      - ./configs/${PRESET:?err}/core/dashd.conf:/dash/.dashcore/dash.conf
+      - ./data/${PRESET:?err}/core:/dash/data
     command:
       - dashd
-      - -conf=/dashd.conf
-      - -datadir=/data
+      - -datadir=/dash/data
```
shumkov: You need to pass option with `CORE_P2P_PORT` too.

docker-compose.yml
```diff
@@ -5,14 +5,13 @@ services:
     image: dashpay/dashd:${CORE_VERSION:-latest}
     restart: always
     ports:
-      - ${CORE_P2P_PORT}:20001 # P2P
+      - ${CORE_P2P_PORT}:${CORE_P2P_PORT} # P2P
```
shumkov: We shouldn't start if var is empty. Please mark it as required: `${CORE_P2P_PORT:?err}`

configs/testnet/core/dashd.conf
```diff
@@ -0,0 +1,30 @@
...
+rpcallowip=172.27.0.0/16
+rpcallowip=192.168.0.0/16
+
+rpcworkqueue=64
+rpcthreads=16
+
+# external network
+listen=1
+bind=0.0.0.0
+port=20001
```
shumkov: Let's remove it from dashd configs since we pass it as an option.

README.md
```diff
@@ -38,7 +39,7 @@ There are two ways to apply a present:
 In order to run a masternode use Docker Compose:
 
 ```bash
-$ docker-compose up
+$ docker-compose up -d
```
shumkov: I think it should up to users do they want deattach process or not.

.env.evonet
shumkov: Let's group `COMPOSE` envs together.

configs/testnet/core/dashd.conf
strophy: Remove just the port, or all options relating to external network?

.env.local
```diff
@@ -6,5 +6,6 @@ MACHINE_VERSION=0.11.0
 DAPI_VERSION=0.11.1
```
shumkov: We need the compose file env var here too

docker-compose.yml
shumkov: I would use `${CORE_P2P_PORT:?err}` here too for consistency

Issue #13, opened 2020-03-20
dash-cli doesn't work in the dashpay container


cloudwheels:
**reproduce**
```
sudo docker exec -it <dashpay_container> /bin/bash
dash-cli help
error: Could not locate RPC credentials. No authentication cookie could be found, and no rpcpassword is set in the configuration file (/dash/.dashcore/dash.conf)
```

**issue**
dash-cli doesn't work in the dashpay container
This may have been done for a reason but I don't understand why.

**suggested fix**
previous versions added following `volume` to dashpay container config in `docker-compose.yml`:
`./data/dash.conf:/root/.dashcore/dash.conf`

strophy:
Workaround: `dash-cli -conf=/dashd.conf <command>`

This occurs because the dash.conf file is not mounted in the default location of `~/.dashcore/dash.conf`. Mounting it to `/dash/.dashcore/dash.conf` (or `/root/.dashcore/dash.conf`) is problematic because the directory does not exist in this location in the image, so bind mounting dash.conf there will create the directory with root owner, but dashd runs as user "dash" in the image and then cannot write the blockchain files to the data folder. The easiest way to fix this is modify the dashpay/dashd image.

shumkov:
@strophy Are you sure it won't work?

strophy:
Yes, I'm fairly sure. We can see it in one of my previous attempts [here](https://github.com/strophy/docker-masternode/blob/master/docker-compose.yml#L10).

This implementation bind-mounts `dash.conf` to `/dash/.dashcore/dash.conf` in the container. If the datadir is not manually specified (so dashd tries to use the default, i.e. `/dash/.dashcore/` then it will encounter the following error on start:

```
dashd       | 2020-03-24 07:36:50 Error: Cannot obtain a lock on data directory /dash/.dashcore. Dash Core is probably already running. No such file or directory.
dashd       | Error: Cannot obtain a lock on data directory /dash/.dashcore. Dash Core is probably already running. No such file or directory.
dashd exited with code 1
```

The `/dash/` directory has the following permissions:
`drwxr-xr-x 2 root root 4.0K Mar 24 07:32 .dashcore`

All other files in the folder have `dash/dash` user and group. If we want to use default locations for these files so `dash-cli` works without passing the `-conf` argument then the directory needs to be created in the docker image with the correct user. Other options are to use an alternative datadir location, and alternative conf file location, or run dashd as root in the container.

strophy:
[...]

PR #18, opened 2020-03-23
fix: dash-cli doesn't work without default config

shumkov:
See details in #13

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,11 +7,11 @@ services:
     ports:
       - 20001:20001 # P2P
     volumes:
-      - ./configs/${PRESET:?err}/core/dashd.conf:/dashd.conf
+      - ./configs/${PRESET:?err}/core/dashd.conf:/dash/.dashcore/dash.conf
       - ./data/${PRESET:?err}/core:/data
     command:
       - dashd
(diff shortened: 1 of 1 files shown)


strophy:
This works because a non-default datadir is specified, but it still creates the bind mount as root user:

```
dash@d9c975a24425:~$ pwd
/dash
dash@d9c975a24425:~$ ls -lha
total 28K
drwxr-xr-x 3 dash dash 4.0K Mar 24 07:55 .
drwxr-xr-x 1 root root 4.0K Mar 24 07:53 ..
-rw------- 1 dash dash  101 Mar 24 07:55 .bash_history
-rw-r--r-- 1 dash dash  220 Aug 31  2015 .bash_logout
-rw-r--r-- 1 dash dash 3.7K Aug 31  2015 .bashrc
drwxr-xr-x 2 root root 4.0K Mar 24 07:53 .dashcore
-rw-r--r-- 1 dash dash  655 May 16  2017 .profile
dash@d9c975a24425:~$
```

shumkov:
@strophy What do you mean non-default? `/data` is `dash` user's home dir.
We put config with root user but we just read data from `/data/.dashcore` dir, so it works.

PR #46, opened 2020-05-11
feat: add a template for issues

thephez:
This will provide an general issue template to prompt for information when someone wants to open a new issue.

### DIFF
--- a/.github/ISSUE_TEMPLATE/issue.md
+++ b/.github/ISSUE_TEMPLATE/issue.md
@@ -1,0 +1,41 @@
+---
+name: Issue
+about: Create a report to help us improve
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+<!--- Provide a general summary of the issue in the Title above -->
+
+## Expected Behavior
+<!--- If you're describing a bug, tell us what should happen -->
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/issue.md
```diff
@@ -0,0 +1,41 @@
+---
```
shumkov, line 1: Can we do labels which are depending on the issue type, like "bug" or "feature"?

thephez: It is possible to define multiple issue templates for different purposes which require users to select an issue type (like [this](https://github.com/thephez/exp/issues/new/choose)). I was trying to keep it simple/generic. We could have 2 options though and still keep it generic enough to use on most repos.

thephez: @shumkov Opened alternative PR #47 with 2 templates that include labels.

shumkov: Awesome! We need more guys to check it before we merge it.

thephez:
Closing this one in favor of #47

Issue #17, opened 2020-03-23
Script to show masternode status


strophy:
Dashman provides a `dashman status` function that shows an overview of the masternode health. 
![image](https://user-images.githubusercontent.com/32928115/77278209-9fc51e00-6d23-11ea-8d93-0481b880c63a.png)

I'd like to implement something similar here, can anyone suggest an appropriate design and language to do this? Then I will make an initial attempt.

shumkov:
Yeah, that's a good idea. There is a lot of L1 and L2 status information that we can output. It would be cool to define what exactly we what to print out for the beginning. Maybe we need several status commands? Or just one with overall info + options that provide more details for a specific section.

cloudwheels:
@strophy NodeJS would make sense for language/platform, built as cli tool (but you could potentially add a web interface)

strophy:
Cool, thanks for kicking this off! I really like the way Docker commands are structured, with general information on the first command and more specific information when further options are provided. How about this for an initial set of commands:

```
mn status           ----> status overview and guide to further commands
mn status core      ----> L1 status, similar to dashman output
mn status platform  ----> L2 status, focus on evo services
mn status network   ----> status of P2P and incoming/outgoing connections
mn status ....      ----> more?
```

This could be followed by further helper commands related to adding config info (`mn config xxx`) and possibly even walking the user through creating and using the registration transactions. Thoughts?

shumkov:
Will `mn status network` be core or platform p2p info?

strophy:
I was thinking of it as similar to the network and peers tab of Dash Core, so L1 status. I don't know enough about what is going on at L2 yet to know what could/should be included from there. Maybe things like status of tendermint consensus could be there too? Input from you, other devs or maybe @thephez would be helpful.

shumkov:
Yeah, I think we need both `mn status core network` and `mn status platform network`

mangus:
I would add:
```
mn status network          --> Both core & platform network status
mn status network core     --> Same as "mn status core network"
mn status network platform --> Same as "mn status platform network"
```

shumkov:
@mangus I guess we will have more `mn status core ...` than just `network`.

strophy:
I'd like to keep the length of commands to 1 or 2 words to keep things simple if possible. How much network information will we realistically be displaying for each one? Maybe we could just include network information in `status core` and `status platform` initially and move it out to a separate command `status network` for both later if we feel it is necessary...

[...]

PR #49, opened 2020-05-14
feat: implement status command

strophy:
The status command currently outputs some information about the host and docker container states. 

I'm unsure about how to properly output this data or if we can share some of the functionality between the `isServiceRunning` and `listContainerStatus` functions. I'd also like to add functionality to output some masternode status by getting various information from inside the containers, and split it up into `mn status host`, `mn status docker`, `mn status masternode` and `mn status all` commands, but thought I would check my progress so far first to make sure I'm going about this the right way.


src/commands/status.js
```diff
@@ -0,0 +1,118 @@
...
+              title: 'Fetch IP info',
+              task: async () => {
+                status.host.ip = await publicip.v4();
+              }
+            },
+          ])
+        )
+      },
+      {
+        title: 'Show Docker status',
```
shumkov, line 84: I would call it "Servies are running"

src/commands/status.js
```diff
@@ -0,0 +1,118 @@
...
+        task: () => (
+          new Listr([
+            {
+              title: 'Show running container states',
```
shumkov, line 88: We could add tasks for each service like "Drive is running"  and use isServiceRunning to check status

src/commands/status.js
```diff
@@ -0,0 +1,118 @@
...
+      docker: {},
+      masternode: {},
+    };
+    const tasks = new Listr([
+      {
+        title: 'Show host status',
+        task: () => (
+          new Listr([
+            {
+              title: 'Fetch hostname',
```
shumkov, line 41: Not sure that they should be a part of Listr. Maybe just use a table to output this data?

src/commands/status.js
strophy: OK, I tried implementing this as a loop on all existing containers to make sure we have visibility on any containers running on the system from other networks or previous attempted installations, since these may help with scenarios like troubleshooting a container stuck in restarting state due to blocked ports. I think using `isServiceRunning` makes the command less useful because we only see the Boolean result of the comparison `(status === 'running')`. This means we lose important details about other possible containers states, e.g. "exited", "restarting", "paused", "dead" etc.

src/commands/status.js
strophy: You mean something like this? https://www.npmjs.com/package/table-layout

shumkov: This one looks better https://www.npmjs.com/package/table

src/commands/status.js
shumkov: I got it. Then I think you can put this info to the table as well and you don't need Listr at all.

strophy: OK, I have re-implemented output using tables, separated the subcommands out to show status for `(host|docker|masternode)` separately, and added a few examples of how I want to execute commands inside the containers. Is this the right approach, or should I move more logic out into functions? It's probably best if we resolve any remaining code style and output issues and close this PR before I continue adding new output.

src/commands/status.js
```diff
@@ -0,0 +1,86 @@
...
+    let output;
+
+    const tableConfig = {
+      //singleLine: true,
+      drawHorizontalLine: (index, size) => {
+        return index === 0 || index === 1 || index === size;
+      }
+    };
+
+    if (targetSystem === 'host') {
```
shumkov, line 37: Maybe it's better to do nested commands instead of params?

src/docker/DockerCompose.js
```diff
@@ -129,6 +129,79 @@ class DockerCompose {
     }
   }
 
+  /**
+   * List all docker compose containers with status
+   *
+   * @param {string} preset
+   * @return {Promise<void>}
+   */
+  async listContainerStatus(preset) {
```
shumkov, line 138: `inspectService(preset, serviceName = undefined)`?

It seems like we could reuse the code from `isServiceRunning`. You could extract duplicate logic to the private method and call it in both methods.

src/commands/status.js
strophy: You mean like `mn status:(host|docker|masternode)` right? I think that makes so much more sense, thanks for the suggestion! Does this look like it is implemented correctly?

shumkov: It looks good! Will have a summary status command `mn status`?

src/commands/status/docker.js
```diff
@@ -0,0 +1,54 @@
...
+
+    const tableConfig = {
+      //singleLine: true,
+      drawHorizontalLine: (index, size) => {
+        return index === 0 || index === 1 || index === size;
+      }
+    };
+
+    data.push(['Container', 'ID', 'Status']);
+    data.push(...(await dockerCompose.listContainerStatus(preset)));
```
shumkov, line 34: Shouldn't we ignore `drive_mongodb_replica_init`? It's not supposed to running and could confuse people? Don't we want to add human-readable service names?

shumkov: I would also use colors for Status column.

strophy: I currently get the service names by inspecting each container returned by the `docker ps -q` command. I can't see any friendly service names in that output, the best I could find were the friendly names we define in the `docker-compose.*.yml` files. Are these friendly enough? Or are you suggesting I convert these names into hard coded names in the CLI?

I'm not sure if I should specifically ignore `drive_mongodb_replica_init` in the command output, maybe we could show when it was last run? What does it actually do, does it only run once or regularly? I think it's better to include it and leave a note on docs.dash.org that it is normal for this service not to be running.

Should I use chalk for colors?

src/commands/status.js
strophy: Ah, now I understand what the oclif documentation means with "topics"! I've added a command showing some basic status information for `mn status`. The nested command classes should still extend `BaseCommand` right?

Let's continue discussion in #17 with more people about what they would like each command in the topic to display? I would add this in smaller future PRs.

src/docker/DockerCompose.js
strophy: Do you mean I should create a private method using the new `#listContainers()` syntax described [here](https://github.com/tc39/proposal-private-methods#private-methods-and-fields)? This would bump our Node version requirement from 10 to 12 I think. Or should I just create a normal method that is not used outside the `DockerCompose` class?

src/commands/status/docker.js
shumkov: Maybe we should just hard code list of services with human-readable names and inspect this list? Or we could have a mapping?

shumkov: the table lib doesn't prove colors for cells? we could use chalk as well.

shumkov: `drive_mongodb_replica_init` starts only to setup replica set and then exit. So it's normal behavior when it exit. Maybe we should only hide it when it exit?

src/docker/DockerCompose.js
shumkov: Yeah, just private method for `DockerCompose`. JS doesn't support private methods so you can just mar it as a private using JSDoc

shumkov: Yep, bump minimal version if you need it.

src/commands/status.js
shumkov: > The nested command classes should still extend BaseCommand right?
yes, I think so.

src/docker/DockerCompose.js
strophy: Actually I got confused since there is a TC39 proposal for true private methods and fields in JS which is supported in Node 12. But I've just followed the style of the other methods in the class for now.

src/commands/status/docker.js
strophy: OK, this is done as well now. I also added human-readable service names, but I don't really think this adds anything useful to the output - it's basically the same information. I have filtered out `drive_mongodb_replica_init` from the results completely, debugging this service can just directly use `docker logs` for advanced users I think?

shumkov: Looks good. I've refactored a bit your code and sent you diff in Slack

strophy: Thanks, this is a huge help to learn better coding practices! What you have done is basically separating the concerns in a more logical way, right? In general, how do you reason about which logic can be extracted into a separate function? Not sure if this is something you can learn or if it just comes from experience.

My only question is about the logic of having the `getContainersList()` function return an array of strings when both `inspectService()` and `isServiceRunning()` will ignore everything except the first string in the array. The previous design of `inspectService()` allowed it to process one or more services at a time. What is the advantage of moving the loop logic into the status command? Should we simplify `getContainersList()` to only return a single container ID instead of an array? Because now it is only ever called together with a `serviceName`, so it only ever returns a single result.

shumkov: > What you have done is basically separating the concerns in a more logical way, right? In general, how do you reason about which logic can be extracted into a separate function?

Usually, if you need to reuse this part of code somewhere else it make sense to extract it to a separate function in order to do not duplicate code.

> My only question is about the logic of having the getContainersList() function return an array of strings when both inspectService() and isServiceRunning() will ignore everything except the first string in the array...

It's possible for a service to have several containers, in case of failure or using the `run` command. I guess we should make `inspectService()` to respond with an array of inspections for the service and handle it in the docker command class.

src/commands/status/docker.js
```diff
@@ -0,0 +1,99 @@
...
+        dapi_insight: 'DAPI Insight',
+        dapi_api: 'DAPI API',
+        dapi_tx_filter_stream: 'DAPI Transactions Filter Stream',
+        dapi_envoy: 'DAPI Envoy',
+        dapi_nginx: 'DAPI Nginx',
+      });
+    }
+
+    const tableRows = [
+      ['Service', 'ID', 'Status'],
```
shumkov, line 44: It would be very useful to add versions here. We could add LABEL to docker images with version and then get it back here via inspect.

src/commands/status/docker.js
```diff
@@ -0,0 +1,99 @@
+const { table } = require('table');
+const chalk = require('chalk');
+
+const ContainerIsNotPresentError = require('../../docker/errors/ContainerIsNotPresentError');
+
+const BaseCommand = require('../../oclif/command/BaseCommand');
+
+const PRESETS = require('../../presets');
+
+class DockerStatusCommand extends BaseCommand {
```
shumkov, line 10: Let’s rename it ”services”

strophy: Done, command is now `mn status:services`

src/commands/status/docker.js
strophy: I've added functionality to support version numbers from labels. The key is read from the `org.dash.version` key, that should be fine right? I've added an [example label to my test image of Sentinel](https://github.com/strophy/docker-sentinel/commit/1aecef0fa02a03f78d389b17632c8871a5ef2c20), does that look right? Should the container track the version of the software it contains? Did you want to add these labels in the Dockerfile or in Docker Compose?

src/commands/status/docker.js
strophy: Sorry for the long delay, I was overloaded with other work. I don't really understand what you mean here, I was thinking more about how the functions pass data around internally rather than any external services with multiple containers. I think it works now, so it's not really important to modify this behaviour until we need it?

src/commands/status/docker.js
shumkov: A container doesn’t need to track it, I suppose. Yeah, let’s just put versions to service Docker files.

strophy: So the container should have a separate version, like 0.1.0? Or same version as Sentinel, like 1.4.0?

src/commands/status/docker.js
shumkov: I mean, practically, it possible that a docker-compose service might have several containers in some cases. For example, in ps output, the first one could be exited due to an error, and the second one created after a service restart. So on your list, you will show just that service is exited but it’s not really true.
We don’t need to handle it right now. Let’s keep it simple :)

src/commands/status/docker.js
shumkov: Yeah, I thought it will be a software version that we run inside a container.

strophy: OK, I'll push and build a version for Sentinel.

shumkov: Let’s do that in the next iteration (separate PR)?

strophy: Versions are already done in 2629b6d - I think we can merge this once conflicts are resolved?

PR #47, opened 2020-05-12
feat: add issue templates for bugs and feature requests

thephez:
Alternative to #46 which only has a single generic issue template. This PR includes separate templates for bug reports and feature requests. The templates automatically apply the `bug` or `enhancement` labels.

PR #52, opened 2020-05-18
chore: add pull request template

thephez:
This provides a PR template to go along with the issue templates added in #47. PR templates are less flexible and I do not see a way to define multiple ones for assigning labels automatically, so I think we are stuck with just having one for now.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,27 @@
+<!--- Provide a general summary of your changes in the Title above -->
+
+## Description
+<!--- Describe your changes in detail -->
+
+## Motivation and Context
+<!--- Why is this change required? What problem does it solve? -->
+<!--- If it fixes an open issue, please link to the issue here. -->
+
+## How Has This Been Tested?
+<!--- Please describe in detail how you tested your changes. -->
+<!--- Include details of your testing environment, and the tests you ran to -->
+<!--- see how your change affects other areas of the code, etc. -->
+
+## Types of changes
+<!--- What types of changes does your code introduce? Put an `x` in all the boxes that apply: -->
+- [ ] Bug fix (non-breaking change which fixes an issue)
+- [ ] New feature (non-breaking change which adds functionality)
+- [ ] Breaking change (fix or feature that would cause existing functionality to change)
+- [ ] Documentation update (if none of the other choices apply)
+
+## Checklist:
+<!--- Go over all the following points, and put an `x` in all the boxes that apply. -->
+- [ ] My code follows the code style of this project.
+- [ ] I have performed a self-review of my own code
+- [ ] I have commented my code, particularly in hard-to-understand areas
+- [ ] I have made corresponding changes to the documentation
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,4 +1,5 @@
 <!--- Provide a general summary of your changes in the Title above -->
+<!--- Pull request titles must use the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) format -->
 
 ## Description
 <!--- Describe your changes in detail -->
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,13 +1,13 @@
 <!--- Provide a general summary of your changes in the Title above -->
 <!--- Pull request titles must use the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) format -->
 
-## Description
-<!--- Describe your changes in detail -->
-
-
-## Motivation and Context
+## Issue being fixed or feature implemented
 <!--- Why is this change required? What problem does it solve? -->
 <!--- If it fixes an open issue, please link to the issue here. -->
+
+
+## What was done?
+<!--- Describe your changes in detail -->
 
 
 ## How Has This Been Tested?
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,27 @@
...
+## Motivation and Context
+<!--- Why is this change required? What problem does it solve? -->
+<!--- If it fixes an open issue, please link to the issue here. -->
+
+## How Has This Been Tested?
+<!--- Please describe in detail how you tested your changes. -->
+<!--- Include details of your testing environment, and the tests you ran to -->
+<!--- see how your change affects other areas of the code, etc. -->
+
+## Types of changes
```
shumkov, line 15: Since we are using https://www.conventionalcommits.org/ we have type in PR title. But I would add to checklist that a developer marked it as breaking change if it breaks compatibility.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,27 @@
...
+<!--- What types of changes does your code introduce? Put an `x` in all the boxes that apply: -->
+- [ ] Bug fix (non-breaking change which fixes an issue)
+- [ ] New feature (non-breaking change which adds functionality)
+- [ ] Breaking change (fix or feature that would cause existing functionality to change)
+- [ ] Documentation update (if none of the other choices apply)
+
+## Checklist:
+<!--- Go over all the following points, and put an `x` in all the boxes that apply. -->
+- [ ] My code follows the code style of this project.
```
shumkov, line 24: We have the linter check via CI, so I suppose we don't need it here. But I would add something like "I have covered by unit/integration/functional/e2e tests"

We could prettify https://docs.google.com/document/d/1DhXios0fvK93LvNXr-WIf9C9kreYsTF5MOLwnJlhkzU/edit and put it in contribution guide and put a reference here.

.github/PULL_REQUEST_TEMPLATE.md
thephez: Do you mean remove these items except for the one that indicates if it is a breaking change?

thephez: We could add this to the Checklist section and remove the "Types of Changes":

>Breaking change:
>- [ ] Yes
>- [ ] No

shumkov: I think we don't need "Types of Changes" coz we specify it in PR title according to https://www.conventionalcommits.org/. Also, if three is breaking change you need to add "!" to PR title as well.
I was thinking to move it to "Checklist" as a reminder.

Probably, it's better to have a separate section "Breaking changes" where a developer should describe them if they are present (we need it for changelog).

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,28 @@
...
+<!--- see how your change affects other areas of the code, etc. -->
+
+## Types of changes
+<!--- What types of changes does your code introduce? Put an `x` in all the boxes that apply: -->
+- [ ] Bug fix (non-breaking change which fixes an issue)
+- [ ] New feature (non-breaking change which adds functionality)
+- [ ] Breaking change (fix or feature that would cause existing functionality to change)
+- [ ] Documentation update (if none of the other choices apply)
+
+## Checklist:
```
shumkov, line 23: Shouldn't we add "I have selected a milestone"?

thephez: Added below

.github/PULL_REQUEST_TEMPLATE.md
thephez: Dropped Types of Changes and added breaking changes section in b24930d

.github/PULL_REQUEST_TEMPLATE.md
thephez: Dropped code style comment in 6598b89

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,32 @@
+<!--- Provide a general summary of your changes in the Title above -->
+<!--- Pull request titles must use the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) format -->
+
+## Description
+<!--- Describe your changes in detail -->
+
+
+## Motivation and Context
```
shumkov, line 8: Shouldn’t it go first?

thephez: I think it does not matter much which is first. I always want to know "what" (the description). The "why" (motivation/context) may be obvious from the description and not need context (or I may not care about it sometimes 🙂). But basically, the template generator I used put them this way and I thought it made sense 😄

shumkov: I just prefer to follow the Agile product-first approach where you describe a problem first and then - solution. In DIPs we do the same ("Motivation"). Also, you see a reference to the issue first where you can get info about that. But you right, it's not so important so it's up to you :).

BTW, is it possible to share these templates among our repos to be able to update them only in one place?

shumkov: Previously we had motivation on the top https://github.com/dashevo/wallet-lib/pull/120

thephez: I changed the heading order and adjusted the wording like the wallet-lib example.

thephez: > BTW, is it possible to share these templates among our repos to be able to update them only in one place?

Unfortunately no :slightly_frowning_face:

PR #96, opened 2020-07-15
refactor: status command output refactoring

strophy:
Minor issues in command output.

## What was done?
Adds a header row to commands where it was missing and fixes help output.

## How Has This Been Tested?
Ran locally

## Breaking Changes
None

## Checklist:

### DIFF
--- a/src/commands/status/host.js
+++ b/src/commands/status/host.js
@@ -11,7 +11,9 @@ class HostStatusCommand extends BaseCommand {
    * @return {Promise<void>}
    */
   async runWithDependencies() {
-    const rows = [];
+    const rows = [
+      ['Key', 'Value'],
+    ];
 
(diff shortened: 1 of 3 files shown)


src/commands/status/host.js
```diff
@@ -11,7 +11,9 @@ class HostStatusCommand extends BaseCommand {
    * @return {Promise<void>}
    */
   async runWithDependencies() {
-    const rows = [];
+    const rows = [
```
shumkov, line 14: Do you think we need such headers? :)) I feel like it's pretty obvious so we can just skip them :)

strophy: Actually I didn't think so, but we are adding a header row to the table so it made sense to put actual headers in it ;) I have removed the header row completely now, is this better?

PR #33, opened 2020-04-25
docs: update README.md to clarify install instructions

cloudwheels:
Docs only updated to address issues #10 and #32 (close on merge)

PR #123, opened 2020-08-19
feat: move container datadirs to named docker volumes

strophy:
This PR moves the `config.json` file managed by the `mn config` command to `~/.mn/config.json` and changes container volume management to named volumes instead of bind mounts. 

This brings the benefit of being able to support multiple configs running on the same network on the same host, and resolves multiple issues relating to container or host users being unable to write directories created by docker as root. [PR #33](https://github.com/dashpay/docker-dashd/pull/33) in docker-dashd must be merged first to support the mountpoint.

## What was done?
- [x] Move config.json from data/config.json to ~/.mn/config.json
- [x] Change datadirs from bind mounts to named volumes
- [x] Change mountpoints to use datadirs writeable by container users

## How Has This Been Tested?
- [x] Creating single testnet/evonet node
- [x] Crashing/deleting/recreating the core container and connecting to the same named volume, recovery works
- [x] Creating evonet node (currently fails due to `Wrong Block.Header.LastResultsHash`)
- [x] Creating multiple testnet nodes on one host
- [ ] Creating multiple evonet nodes on one host (currently fails due to port conflicts)
- [ ] Test triggering error if config datadir cannot be created

## Breaking Changes
- [ ] Tendermint cleanup process in `mn reset` probably needs updating
- [ ] Does any other part of mn-bootstrap expect to be able to access datadirs on the host filesystem as a bind mount?
- [ ] Has [PR #33](https://github.com/dashpay/docker-dashd/pull/33) been merged?

## Checklist:


docker-compose.yml
```diff
@@ -30,3 +28,6 @@ services:
       - RPCPORT=20002
       - NETWORK=${NETWORK?:err}
       - SENTINEL_ARGS=-b
+
+volumes: 
+  core:
```
shumkov, line 33: core_data ?

docker-compose.platform.yml
```diff
@@ -131,3 +131,4 @@ services:
 volumes:
   drive_mongodb:
   drive_leveldb:
+  drive_tendermint:
```
shumkov, line 134: drive_tendermint_data?

docker-compose.platform.yml
```diff
@@ -131,3 +131,4 @@ services:
 volumes:
   drive_mongodb:
```
shumkov, line 132: drive_mongodb_data

src/commands/reset.js
```diff
@@ -46,6 +46,7 @@ class ResetCommand extends BaseCommand {
             {
               title: 'Cleanup Core data dir',
```
shumkov, line 47: Since we store data in volumes we don't need to remove dirs anymore and everything will be done by "Remove Docker containers and associated data" task

src/config/configFile/dataDirRepository.js
```diff
@@ -0,0 +1,26 @@
+const fs = require('fs');
+const path = require('path');
+const os = require('os');
+
+const CouldNotCreateDatadirError = require('../errors/CouldNotCreateDatadirError');
+
+class DataDirRepository {
```
shumkov, line 7: I would call it home dir. But nor sure that repository pattern fits here. A repository supposes to store/fetch your models. I would just make a function `ensureHomeDir`.

src/createDIContainer.js
```diff
@@ -53,7 +55,8 @@ async function createDIContainer() {
    * Config
    */
   container.register({
-    configFilePath: asValue(path.resolve(__dirname, '../data/config.json')),
+    dataDirRepository: asClass(dataDirRepository),
+    configFilePath: asValue(path.resolve(os.homedir(), '.mn/config.json')),
```
shumkov, line 59: I think we don't need it since we have `homeDirPath`. I would allow people to connfigure home dir path with `MN_HOME_DIR_PATH` env variable.

docker-compose.platform.yml
```diff
@@ -45,7 +45,7 @@ services:
       - 26656:26656 # P2P
     volumes:
       - ./configs/${NETWORK:?err}/drive/tendermint:/tendermint/config
```
shumkov, line 47: You can't mount file into already mounted dir (`drive_tendermint:/tendermint`) so I would bring back the previous separation of data and config dirs. To do so you need to revert tendermint config and mount data dir to named volume (`drive_tendermint_data:/data`)

src/config/Config.js
```diff
@@ -119,6 +119,7 @@ class Config {
   toEnvs() {
     return {
       COMPOSE_PROJECT_NAME: `dash_masternode_${this.getName()}`,
+      CONFIG_NAME: this.getName(),
```
shumkov, line 122: Do we still need it?

docker-compose.platform.yml
strophy: Actually this works fine, in the same way as it now works for Dash Core if using the correctly prepared image. All you need to do is make sure the dirs exist in the image with the correct owner before mounting anything. It looks like this inside the running container:
```
bash-4.4$ ls -lha
total 32
drwxr-sr-x    5 tmuser   tmuser      4.0K Aug 25 07:49 .
drwxr-xr-x    1 root     root        4.0K Aug 25 07:49 ..
drwxrwxr-x    2 1000     tmuser      4.0K Aug 25 07:14 config
drwx--S---    2 tmuser   tmuser      4.0K Aug 25 07:49 data
drwxr-sr-x    4 tmuser   tmuser      4.0K Aug 25 07:49 db
-rw-------    1 tmuser   tmuser       148 Aug 25 07:49 node_key.json
-rw-------    1 tmuser   tmuser       345 Aug 25 07:49 priv_validator_key.json
-rw-------    1 tmuser   tmuser        48 Aug 25 07:49 priv_validator_state.json
bash-4.4$ ls config -lha
total 40
drwxrwxr-x    2 1000     tmuser      4.0K Aug 25 07:14 .
drwxr-sr-x    5 tmuser   tmuser      4.0K Aug 25 07:49 ..
-rw-rw-r--    1 1000     tmuser     13.5K Aug 25 07:14 config.toml
-rw-rw-r--    1 1000     tmuser     13.5K Aug 21 00:01 genesis.json
bash-4.4$ 
```

shumkov: Excellent! 👍

src/config/configFile/dataDirRepository.js
strophy: OK, I tried to refactor this into a function. Still unsure about error handling here.

src/createDIContainer.js
strophy: You mean for running standalone docker-compose without the CLI? Why does it need access to ~/.mn/config.json, it can't read this file, all config is taken from envs. Don't really understand what you mean here.

docker-compose.platform.yml
```diff
@@ -129,5 +129,6 @@ services:
       - ./configs/${NETWORK:?err}/dapi/nginx/errors.grpc_conf:/etc/nginx/conf.d/errors.grpc_conf
 
 volumes:
-  drive_mongodb:
-  drive_leveldb:
+  drive_mongodb_data:
+  drive_leveldb_data:
```
shumkov, line 133: It's more like `drive_abci_leveldb`. Isn't it? So we are prefixing with a service name.

docker-compose.platform.yml
shumkov: Are you sure those configs are the same? :))

configs/evonet/drive/tendermint/config.toml
```diff
@@ -50,17 +50,17 @@ log_format = "plain"
 genesis_file = "config/genesis.json"
 
 # Path to the JSON file containing the private key to use as a validator in the consensus protocol
-priv_validator_key_file = "/data/priv_validator_key.json"
+priv_validator_key_file = "/tendermint/priv_validator_key.json"
```
shumkov, line 53: What if we just make all paths as in default tendermint config and mount `$HOME/.tendermint` to volume?

```
# NOTE: Any path below can be absolute (e.g. "/var/myawesomeapp/data") or
# relative to the home directory (e.g. "data"). The home directory is
# "$HOME/.tendermint" by default, but could be changed via $TMHOME env variable
# or --home cmd flag.
```

src/config/configFile/ensureHomeDirFactory.js
```diff
@@ -0,0 +1,30 @@
...
+const path = require('path');
+const os = require('os');
+
+const CouldNotCreateHomeDirError = require('../errors/CouldNotCreateHomeDirError')
+
+/**
+ * @param {string} homeDirPath
+ * @return {ensureHomeDir}
+ */
+function ensureHomeDirFactory(homeDirPath = path.resolve(os.homedir(), '.mn')) {
```
shumkov, line 11: I would remove the default value for the dependency. It should be defined in the container.

src/config/configFile/ensureHomeDirFactory.js
```diff
@@ -0,0 +1,30 @@
...
+  /**
+   * @typedef {ensureHomeDir}
+   * @return {string} homeDirPath
+   */
+  function ensureHomeDir() {
+    if (!fs.existsSync(homeDirPath)) {
```
shumkov, line 17: If you just check existence in return then it will make this function more readable:

```suggestion
    if (fs.existsSync(homeDirPath)) {
       return homeDirPath;
    }
    
    ...
```

src/oclif/command/BaseCommand.js
```diff
@@ -17,6 +17,16 @@ class BaseCommand extends Command {
...
     this.container = await createDIContainer();
 
+    // Set up config dir
+    /**
+     * @type {ensureHomeDir}
+     */
+    const ensureHomeDir = this.container.resolve('ensureHomeDir');
+    const homeDirPath = ensureHomeDir();
+    this.container.register({
+      configFilePath: asValue(homeDirPath + '/config.json'),
```
shumkov, line 27: This should be in the container since we already know `homeDirPath` there.

configs/evonet/drive/tendermint/config.toml
strophy: - Mounting the volume at `$HOME/.tendermint` doesn't work because upstream [defines](https://github.com/tendermint/tendermint/blob/master/DOCKER/Dockerfile#L9) `ENV TMHOME /tendermint` so we should mount the docker volume here to store all mutable state.
- Inside this, it works better to bind mount the dir instead of individual files to avoid implicitly creating the `config` dir as root inside the container. Instead, it will take the uid:gid of the host system. We shouldn't write anything to this dir or it will appear at the path `configs/${NETWORK:?err}/drive/tendermint`.
- For this reason, not all paths can be set to default, because 3 files (`priv_validator_key.json`, `node_key.json`, `addrbook.json`) are generated on init and therefore mutable. They should live in the `data` dir so they are cleaned up when the volume is removed.

If we bind mount the `config.toml` and `genesis.json` files then `config` owner is root and the 3 files cannot be written, resulting in container crash.

src/oclif/command/BaseCommand.js
strophy: I think I am not understanding awilix again here. Didn't you tell me [here](https://github.com/dashevo/mn-bootstrap/pull/123#pullrequestreview-470333267) to move this out of the container? We can't know the `homeDirPath` before running `ensureHomeDir()`, and we can't run the function before we get to init. So I don't know how we can register it on container creation, I think it need to be registered later? Also not clear how the user can specify a custom homedir here, where can we pass in and read the argument or env?

src/config/configFile/ensureHomeDirFactory.js
strophy: Done, is that what you mean?

shumkov (approved):
Good job, mate! 👍

PR #124, opened 2020-08-20
feat: more status command output

strophy:
Adds more output to the status command

Monitor dmn status, tendermint status and service versions

## What was done?
- [x] Track more layer 1 status info
- [x] Track layer 2 status info
- [x] Track service versions

## How Has This Been Tested?
- [x] Test output on unregistered mn
- [x] Test output on registered mn
- [x] Test output on banned mn

## Checklist:


strophy:
@shumkov do you think we can get version numbers for `status:services` from the mn-bootstrap Config object now? Or do you want to implement `org.dash.version` labels in each image like the original approach [here](https://github.com/dashevo/mn-bootstrap/blob/v0.15-dev/src/commands/status/services.js#L58-L60)?

src/commands/status/masternode.js
```diff
@@ -19,23 +20,57 @@ class MasternodeStatusCommand extends BaseCommand {
...
+    const dashdVersion = (await dockerCompose.execCommand(
       config.toEnvs(),
       'core',
       'dashd --version',
-    );
+    )).out.split('\n')[0];
 
-    rows.push(['Version', versionOutput.out.split('\n')[0]]);
+    // Sync status
+    const mnsyncStatus = JSON.parse((await dockerCompose.execCommand(
```
shumkov, line 30: We can use RPC instead of calling through docker which is much slower. You can find examples in the project code.

shumkov:
@strophy Yeah, I would add an "image" column and print ${docker.image}:${version} there.

src/commands/status/masternode.js
strophy: Like [this](https://github.com/dashevo/mn-bootstrap/pull/124/commits/761505be33b6f042ee8ee7698ae5e553b1012572)? Is something similar available for the Tendermint calls?

strophy:
> @strophy Yeah, I would add an "image" column and print ${docker.image}:${version} there.

Like [this](https://github.com/dashevo/mn-bootstrap/pull/124/commits/1374d9dd3436ee51b5fb2b6304673d6365b771b0)? I don't understand the syntax you are using, this code I have written looks too ugly so I assume you mean something else... Also, should we move the sentinel image into the config?

docker-compose.yml
```diff
@@ -6,6 +6,7 @@ services:
     restart: unless-stopped
     ports:
       - ${CORE_P2P_PORT:?err}:${CORE_P2P_PORT:?err} # P2P
+      - 20002:20002
```
shumkov, line 9: I wouldn't open RPC ports by default. It's pretty dangerous.

src/commands/status/masternode.js
```diff
@@ -1,40 +1,84 @@
...
+    rows.push(['Masternode Status', masternodeStatus.state]);
+    if (masternodeStatus.state === 'READY') {
+      rows.push(['ProTx Hash', masternodeStatus.proTxHash]);
+      rows.push(['Service', masternodeStatus.dmnState.service]);
+      rows.push(['PoSe Penalty', masternodeStatus.dmnState.PoSePenalty]);
+    }
+    rows.push(['Sentinel', (sentinelState !== '' ? sentinelState : 'No errors')]);
+    rows.push(['Port Check', config.options.core.p2p.port + ' ' + portState]);
+    if (config.network !== 'testnet') {
+      rows.push(['Tendermint Version', tendermintVersion]);
```
shumkov, line 78: I would add latest app hash too

src/commands/status/masternode.js
```diff
@@ -1,40 +1,84 @@
...
+    rows.push(['Headers', blockchainInfo.headers]);
+    rows.push(['Blocks', blockchainInfo.blocks]);
+    rows.push(['Masternode Status', masternodeStatus.state]);
+    if (masternodeStatus.state === 'READY') {
+      rows.push(['ProTx Hash', masternodeStatus.proTxHash]);
+      rows.push(['Service', masternodeStatus.dmnState.service]);
+      rows.push(['PoSe Penalty', masternodeStatus.dmnState.PoSePenalty]);
+    }
+    rows.push(['Sentinel', (sentinelState !== '' ? sentinelState : 'No errors')]);
+    rows.push(['Port Check', config.options.core.p2p.port + ' ' + portState]);
```
shumkov, line 76: Can we do port check for Tendermint too?

src/commands/status/services.js
```diff
@@ -77,9 +71,26 @@ class ServicesStatusCommand extends BaseCommand {
         statusText = chalk.keyword(status === 'running' ? 'green' : 'red')(status);
       }
 
+      let path;
+      let image;
+      let version;
+
+      if (serviceName === 'sentinel') { continue };
+      if (serviceName === 'dapi_tx_filter_stream') { continue };
```
shumkov, line 79: it uses the same version as a dapi api

src/commands/status/masternode.js
shumkov: Tendermint has RPC too, but in this case, we need to publish port such as for Core. We really don't want to do that. I didn't take it into account. Probably, it's better to use docker exec for now.

strophy: OK, I have changed it back. Doesn't the `rpcallowip` we define restrict which hosts could access RPC? Maybe if we just define it more restrictively by removing (at least) `rpcallowip=192.168.0.0/16` it would be safer. Could we also randomly generate a different RPC password on startup each time?

Another solution I looked into to make the delay less noticeable was using [table streaming](https://github.com/gajus/table#streaming) but this output has a bug that doesn't support single line mode, so I don't like the appearance much.

Would it be possible to use `Promise.all()` to run the requests asynchronously? This would also provide a speedup by running all the `docker exec` commands simultaneously right?

src/commands/status/masternode.js
strophy: Yes, but I think we need to change the config schema so it contains the host port for Tendermint P2P first, otherwise the port will be hardcoded here like I have done now. This is also preventing us from running multiple evonet nodes on one host at the moment. Should I try a PR for this?

src/commands/status/masternode.js
strophy: Is app hash the hex code after the version? It is already included in the output I think: `0.32.12-f11eae07`

src/commands/status/masternode.js
shumkov: `rpcallowip=192.168.0.0/16` this needed for Docker, we can't remove it, but in general, I agree with you. Also, I agree that we need to generate Core RPC user and password and store it in an MN config.

shumkov: With `Promise.all` it should be better.

shumkov: Can configure Tendermint RPC for local requests as well? https://docs.tendermint.com/master/rpc/

src/commands/status/masternode.js
shumkov: Yep, we can do that ;)

src/commands/status/masternode.js
strophy: OK, added. But this is always blank for me on fully synced evonet-5, is that expected?

src/commands/status/masternode.js
strophy: I have implemented it to read the tendermint port from docker inspect instead, so we can support both current and future tendermint config variants.

src/commands/status/services.js
```diff
@@ -39,13 +39,14 @@ class ServicesStatusCommand extends BaseCommand {
     }
 
     const tableRows = [
-      ['Service', 'Container ID', 'Version', 'Status'],
+      ['Service', 'Container ID', 'Image', 'Version', 'Status'],
```
shumkov, line 42: It's not actually version, it's a tag. It could be anything. I would just keep them together in the image column. 

P.S. Probably we should combine them in configs as well.

strophy: OK, done. Let's change the configs in a separate PR together with the networking fixes.

shumkov: I would separate them into two to make them more atomic

shumkov (approved):
👍

strophy:
Picking this up again now dependencies for RPC access have been merged. I suggest modifying output as follows:

Command intended to show most common status problems at a glance, with green/red colour highlights
```
mn status
- masternode (true/false)
- masternode status
- core sync (true/false)
- tenderdash sync (true/false)
- sentinel status
- protx hash
- service address:port
- PoSe penalty
- last paid
- payment queue position
```

Command intended to help troubleshoot hosting problems, e.g. out of memory/disk, insufficient performance
```
mn status:host
- hostname
- uptime
- platform
- arch
- username
- loadavg
- diskfree
- memory
- CPUs
- IP
```

Command intended to help check status of Docker services, with green/red colour highlights
```
mn status:services
- list of services and state (name, ID, image:tag, status)
```

Command intended for in-depth troubleshooting of node status
```
mn status:node
- masternode
  - isMasternode (true/false)
  - masternode state
  - masternode status
  - mnsync status
  - PoSe penalty
  - lastpaid
  - payment queue position
  - collateral address balance
- core
  - version
  - compare version to latest github release
  - sync (true/false)
  - sync assetname
  - port check
  - header height
  - block height
  - block height (remote explorer)
  - peer count
  - difficulty
- sentinel
  - version
  - status
- tenderdash
  - version
  - port check
  - block height
  - peer count
  - sync (true/false)
  - app hash
```

shumkov:
Sounds good, but I would break the node to `core` and `platform` and make `mn status` as a summary of all of them to represent that everything is working properly.

How about this?

- status:masternode
   - masternode status - disabled/syncing/running - would be nice to show sync progress in %
   - sentinel status
   - protx hash
   - PoSe penalty
   - last paid
   - payment queue position
- status:core
   - version
   - compare version to latest github release
   - network - local / devnet (name) / testnet / mainnet
   - status: not started / failed / syncing / running - would be nice to show sync progress in %
   - sync assetname
   - peer count
   - p2p IP:port
   - p2p port check
   - RPC host:port
   - header height
   - block height
   - network block height (remote explorer)
   - difficulty
   - sentinel version
   - sentinel status
- status:platform
   - version
   - network - local / devnet (name) / testnet / mainnet
   - status - not started / failed / syncing / running - would be nice to show sync progress in %
   - block height
   - network block height (remote explorer)
   - peer count
   - app hash
   - HTTP IP:port
   - HTTP port check
   - gRPC IP:port
   - gRPC port check
   - p2p IP:port
   - p2p port check
   - RPC host:port
- status:host
   - ...
- status:services
  - ...
- status
   - network
   - masternode status
   - core status
   - core version
   - platform status
   - platform version
   - PoSe penalty
   - last paid
   - payment queue position

riongull:
I like the idea of making the `mn status` command bring all other statuses together

strophy:
OK @shumkov this has been reimplemented according to the output we specified above. We could still add some colour coding with chalk I think? Should some of these functions (like figuring out the payment queue position) be factored out into separate files to avoid repetition?

src/commands/status/core.js
```diff
@@ -0,0 +1,114 @@
...
+          port: config.get('core.rpc.port'),
+          user: config.get('core.rpc.user'),
+          pass: config.get('core.rpc.password'),
+        },
+      ),
+      dockerCompose.docker.getContainer('core'),
+    );
+
+    let insightUrl;
+    switch (config.options.network) {
```
shumkov, line 39: I would recommend to define an object with key as network name and value as insight URL so you can do just `insightURLs[networkName]` to get specific URL. It will look cleaner than switch.

src/commands/status/core.js
```diff
@@ -0,0 +1,114 @@
...
+      case 'evonet': insightUrl = 'http://insight.evonet.networks.dash.org:3001/insight-api'; break;
+      case 'testnet': insightUrl = 'https://testnet-insight.dashevo.org/insight-api'; break;
+      case 'mainnet': insightUrl = 'https://insight.dash.org/insight-api'; break;
+      default: insightUrl = 'https://insight.dash.org/insight-api';
+    }
+
+    // Collect data
+    const blockchainInfo = (await coreService.getRpcClient().getBlockchainInfo()).result;
```
shumkov, line 47: You can use object destruction to make code cleaner: `const { result: blockchainInfo } = await ...`

src/commands/status/core.js
```diff
@@ -0,0 +1,114 @@
...
+    const networkInfo = (await coreService.getRpcClient().getNetworkInfo()).result;
+    const mnsyncStatus = (await coreService.getRpcClient().mnsync('status')).result;
+    const peerInfo = (await coreService.getRpcClient().getPeerInfo()).result;
+    const latestVersion = JSON.parse(await fetch('https://api.github.com/repos/dashpay/dash/releases/latest').then((res) => res.text()));
```
shumkov, line 51: You can use object destruction to set a specific property to a variable.

src/commands/status/core.js
strophy: Thanks, these little style hints are incredibly useful.

src/commands/status/core.js
shumkov: Usually, it’s better to put execution rest to variable and pass it to the next function. It’s much more readable than oneliners. Also, I would recommend to do not combine the await and promise-like approach. This is considered smelling according to best practices.

src/commands/status/core.js
```diff
@@ -0,0 +1,152 @@
...
+    };
+
+    // Collect data
+    const { result: blockchainInfo } = await coreService.getRpcClient().getBlockchainInfo();
+    const { result: networkInfo } = await coreService.getRpcClient().getNetworkInfo();
+    const { result: mnsyncStatus } = await coreService.getRpcClient().mnsync('status');
+    const { result: peerInfo } = await coreService.getRpcClient().getPeerInfo();
+    const latestVersionRes = await fetch('https://api.github.com/repos/dashpay/dash/releases/latest');
+    const latestVersion = await latestVersionRes.json();
+    const insightBlockHeightRes = await fetch(`${insightURLs[config.options.network]}/status`);
```
shumkov, line 52: For local network, it will be invalid url. Let’s skip it if the `!insightURL[..`

src/commands/status/core.js
```diff
@@ -0,0 +1,152 @@
...
+    if (status === 'running') {
+      if (mnsyncStatus.AssetName !== 'MASTERNODE_SYNC_FINISHED') {
+        status = `syncing ${(blockchainInfo.verificationprogress * 100).toFixed(2)}%`;
+      }
+    }
+
+    // Apply colors
+    if (status === 'running') {
+      status = chalk.green(status);
+    } else if (status.includes('syncing')) {
```
shumkov, line 90: startsWith

src/commands/status/core.js
```diff
@@ -0,0 +1,152 @@
...
+        },
+      } = await dockerCompose.inspectService(config.toEnvs(), 'core'));
+    } catch (e) {
+      if (e instanceof ContainerIsNotPresentError) {
+        status = 'not started';
+      }
+    }
+    if (status === 'running') {
+      if (mnsyncStatus.AssetName !== 'MASTERNODE_SYNC_FINISHED') {
+        status = `syncing ${(blockchainInfo.verificationprogress * 100).toFixed(2)}%`;
```
shumkov, line 83: Does verification progress include blocks syncing?

src/commands/status/core.js
```diff
@@ -0,0 +1,152 @@
...
+    if (corePortState === 'OPEN') {
+      corePortState = chalk.green(corePortState);
+    } else {
+      corePortState = chalk.red(corePortState);
+    }
+
+    let blocks;
+    if (blockchainInfo.blocks === blockchainInfo.headers
+      || blockchainInfo.blocks >= insightBlockHeight.info.blocks) {
+      blocks = chalk.green(blockchainInfo.blocks);
```
shumkov, line 111: I guess it could be yellow if it’s 1-2 blocks behind?

src/commands/status/core.js
```diff
@@ -0,0 +1,152 @@
...
+      status = chalk.green(status);
+    } else if (status.includes('syncing')) {
+      status = chalk.yellow(status);
+    } else {
+      status = chalk.red(status);
+    }
+
+    if (coreVersion.substring(coreVersion.indexOf(':') + 1) === latestVersion.tag_name.substring(1)) {
+      coreVersion = chalk.green(coreVersion);
+    } else {
```
shumkov, line 98: Could be yellow if second segment is the same (i.e. 0.16)

src/commands/status/index.js
```diff
@@ -0,0 +1,217 @@
...
+            - coreBlocks;
+        } else if (masternodeLastPaidHeight === 0) {
+          paymentQueuePosition = masternodeRegisteredHeight
+            + masternodeEnabledCount
+            - coreBlocks;
+        } else {
+          paymentQueuePosition = masternodeLastPaidHeight
+            + masternodeEnabledCount
+            - coreBlocks;
+        }
```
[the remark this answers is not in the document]
strophy: Moved to a function in the `util` directory.

src/commands/status/platform.js
```diff
@@ -0,0 +1,176 @@
...
+    } = await tendermintStatusRes.json();
+
+    const tendermintNetInfoRes = await fetch(`http://localhost:${config.options.platform.drive.tendermint.rpc.port}/net_info`);
+    const {
+      result: {
+        n_peers: platformPeers,
+      },
+    } = await tendermintNetInfoRes.json();
+
+    const explorerBlockHeightRes = await fetch('https://rpc.cloudwheels.net:26657/status');
```
shumkov, line 82: Let's check it if we use network evonet.

strophy: I've fixed this by following the same pattern as in `status:core` to check if an external block explorer exists for the current config before running the query. This condition will currently never be reached, since there is a testnet check/abort at the top of the command and the command can never run for `local` config because core never reaches `IsSynced` status. But this check will still be useful as platform is enabled on more networks.

shumkov (approved):
Looks good! 👍

PR #134, opened 2020-08-27
feat: nginx responds with unimplemented in case if version is not supported

shuplenkov:

Nginx should handle API requests for undefined versions. * To do so we need to add a location with wildcard instead of version and respond with UNIMPLEMENTED (12) gRPC status code.

## What was done?

Changed nginx configs

## How Has This Been Tested?

Manually

## Breaking Changes

None

## Checklist:

### DIFF
--- a/configs/evonet/dapi/nginx/default.conf
+++ b/configs/evonet/dapi/nginx/default.conf
@@ -39,7 +39,7 @@ server {
     }
 
     # Route GRPC-WEB requests to Envoy DAPI TransactionsFilterStream service
-    location /org.dash.platform.dapi.v0.Core/subscribeToTransactionsWithProofs {
+    location = /org.dash.platform.dapi.v0.Core/subscribeToTransactionsWithProofs {
         if ($request_method = 'OPTIONS') {
             add_header 'Access-Control-Allow-Origin' '*' always;
             add_header 'Access-Control-Allow-Methods' 'GET, POST, OPTIONS' always;
@@ -67,7 +67,7 @@ server {
     }
 
     # Route GRPC-WEB requests to Envoy DAPI API service
-    location ~ /org.dash.platform.dapi.v0.Core {
+    location ~ ^/org\.dash\.platform\.dapi\.v0\.Core {
         if ($request_method = 'OPTIONS') {
             add_header 'Access-Control-Allow-Origin' '*' always;
             add_header 'Access-Control-Allow-Methods' 'GET, POST, OPTIONS' always;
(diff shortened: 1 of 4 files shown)


configs/evonet/dapi/nginx/default.conf
```diff
@@ -67,7 +67,7 @@ server {
     }
 
     # Route GRPC-WEB requests to Envoy DAPI API service
-    location /org.dash.platform.dapi.v0.Core {
+    location ~ /org.dash.platform.dapi.v0.Core {
```
shumkov, line 70: Why do we need regexp modifier here?

shumkov (changes requested):

configs/evonet/dapi/nginx/default.conf
```diff
@@ -149,4 +149,37 @@ server {
...
     }
+
+    location ~ /org.dash.platform.dapi.v[0-9]+.Core {
+        add_header 'Access-Control-Allow-Origin' '*' always;
+        add_header 'Access-Control-Allow-Methods' 'GET, POST, OPTIONS' always;
+        add_header 'Access-Control-Allow-Headers' 'DNT,X-CustomHeader,Keep-Alive,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Content-Transfer-Encoding,Custom-Header-1,X-Accept-Content-Transfer-Encoding,X-Accept-Response-Streaming,X-User-Agent,X-Grpc-Web' always;
+        add_header 'Access-Control-Expose-Headers' 'Content-Transfer-Encoding,Grpc-Message,Grpc-Status,Custom-Header-1' always;
+        add_header 'Content-Type' 'application/grpc-web+proto' always;
+        add_header 'grpc-status' 12 always;
+        add_header 'grpc-message' 'Specified protocol version is not supported' always;
```
shumkov, line 160: protocol -> service

configs/local/dapi/nginx/grpc.conf
```diff
@@ -20,20 +20,41 @@ server {
...
+    }
+
+    location ~ /org.dash.platform.dapi.v[0-9]+.Platform {
+      add_header content-type application/grpc;
+      add_header grpc-status 12;
+      add_header grpc-message "Specified protocol version is not supported";
+      return 204;
+   }
+
+    location ~ /org.dash.platform.dapi.v[0-9]+.Health {
```
shumkov, line 51: do we have such service?

configs/evonet/dapi/nginx/default.conf
shuplenkov: We'll get into the `/org.dash.platform.dapi.v[0-9]+.Core` location without it.

configs/local/dapi/nginx/grpc.conf
shuplenkov: We have such location

configs/evonet/dapi/nginx/default.conf
shumkov: Should we then escape dots in regexp?

shuplenkov: it just works

shumkov: Right, due to dot in regexp represents any char. So it will work if a requested service name will be something like `/org1dash2platform3dapi4v56Core`. We shouldn't allow that I guess.

shumkov (approved):
👍

PR #140, opened 2020-08-31
fix: can't generate genesis config for local

shumkov:

For the local network, we do not provide `genesis.json` which should be dynamically generated by `tendermint int` command from docker compose config. So we need to move this file to tendermint data volume to make it writeable for tendermint.

## What was done?

`genesis.json` is set to writable `data` dir.

## How Has This Been Tested?

With running `mn setup-for-local-development`

## Breaking Changes

None

## Checklist:


antouhou (approved):
👍

PR #142, opened 2020-09-01
fix: set config bind mount read-only

strophy:
Set `config` dir read only to prevent accidentally writing mutable state.

It is not possible to bind-mount the config files directly in `docker-compose.platform.yml` because not all configs provide a `genesis.json` file. Also the `~/config` directory does not exist in the Docker image, which would result in creation of the config dir as root. However, it is still important not to write mutable data to this directory. This PR ensures that the config dir is read-only.

## What was done?
Add `:ro` flag to bind mount instruction.

## How Has This Been Tested?
Started masternode and attempted to reproduce the bug fixed in #140 . Masternode startup fails with logs showing the error `ERROR: open /tendermint/config/genesis.json: read-only file system`

## Checklist:

### DIFF
--- a/docker-compose.platform.yml
+++ b/docker-compose.platform.yml
@@ -44,7 +44,7 @@ services:
     ports:
       - 26656:26656 # P2P
     volumes:
-      - ./configs/${NETWORK:?err}/drive/tendermint:/tendermint/config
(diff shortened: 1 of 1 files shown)


shumkov (approved):
👍

PR #149, opened 2020-09-08
fix: services.core.ports contains an invalid type

strophy:
Removes quotes from generated .env file, since these are not handled by docker-compose.

Trying to start a node with `mn config:envs -o .env && docker-compose up` fails with various errors as follows:
```
services.core.ports contains an invalid type, it should be a number, or an object
ERROR: .FileNotFoundError: [Errno 2] No such file or directory: './"docker-compose.yml"'
```

According to the [env file spec](https://docs.docker.com/compose/env-file/):
- There is no special handling of quotation marks. This means that **they are part of the VAL.**

## What was done?
Remove quotes from generated output.

## How Has This Been Tested?
`mn config:envs -o .env && docker-compose up`
  --> no errors, services start as expected

## Checklist:

### DIFF
--- a/src/commands/config/envs.js
+++ b/src/commands/config/envs.js
@@ -22,7 +22,7 @@ class ConfigEnvsCommand extends BaseCommand {
     let envOutput = '';
 
     for (const [key, value] of Object.entries(config.toEnvs())) {
-      envOutput += `${key}="${value}"\n`;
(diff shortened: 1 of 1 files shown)


shumkov (approved):
Nice catch! 👍

PR #150, opened 2020-09-16
feat: add network parameters to configs

strophy:
This PR adds port parameters to the config to allow access to core RPC from the Docker host and remove hard-coded port numbers from the codebase, making it possible to run more than one mn of the same network on the host without running into Docker port binding conflicts. 

This change is required to refactor #124 for better performance and to be able to run more than one masternode of the same network on the docker host.

## What was done?
- [x] Tightened core RPC security by only accepting connections from Class B & C private networks used by Docker bridge (note Docker Swarm uses Class A)
- [x] Defined rule to accept RPC connections from Docker host (NodeJS)
- [x] Added core RPC port to config schema
- [x] Add core ZMQ port to config
- [x] Add platform RPC ports to config
- [ ] Other platform ports need to be defined (or can platform P2P use dynamic ports?)
- [ ] Would it be easier to start templating the config presets in `/configs/<network>/` instead of passing all this around as envs?

## How Has This Been Tested?
- [x] Tested `curl` from docker host to core container and received output
- [x] Tested `curl` from another host on the local network and received `Connection refused`
- [x] Tested masternode setup on testnet and evonet
- [x] Tested `register` command on evonet
- [ ] Tested `setup-for-local-development`, core steps work but platform init fails with `Failed to perform standard injections with reason: Plugin ChainPlugin of type Standard onStart failed: No available addresses` - what is causing this?
- [ ] Test ZMQ using port defined in config
- [ ] Test platform containers on other ports

## Breaking Changes
- [ ] Insight cannot connect if not using port 20002, templating solution needed

## Checklist:

- [ ] I have performed a self-review of my own code


configs/local/core/dashd.conf
```diff
@@ -20,12 +20,9 @@ zmqpubhashblock=tcp://0.0.0.0:29998
 
 # JSONRPC
 server=1
-rpcport=20002
 rpcuser=dashrpc
 rpcpassword=password
 
-rpcallowip=172.27.0.0/16
-rpcallowip=127.0.0.0
```
shumkov, line 28: We do not allow requests from the local machine?

src/networks.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
   LOCAL: 'local',
-  EVONET: 'evonet',
+  EVONET: 'evonet-6',
```
shumkov, line 3: It should stay `evonet` otherwise other things won't work.

src/templates/renderServiceTemplates.js
```diff
@@ -0,0 +1,27 @@
...
+const crypto = require('crypto');
+
+/**
+ * Render templates for services
+ * @param {Config} config
+ * @returns {Promise<void>}
+ */
+async function renderServiceTemplates(config) {
+  const rpcpassword = crypto.randomBytes(12).toString('base64');
+  try {
```
shumkov, line 12: What is the point of global try/catch?

src/templates/renderServiceTemplates.js
```diff
@@ -0,0 +1,27 @@
...
+    const files = fs.readdirSync('./templates');
+    console.log(config);
+    for (const file of files) {
+      console.log(file);
+      const fileContents = fs.readFileSync('./templates/' + file, 'utf-8');
+      const fileTemplate = dots.template(fileContents);
+      const fileOutput = fileTemplate({config, rpcpassword});
+      console.log(fileOutput);
```
shumkov, line 20: I guess this function should respond with a list of files and their content (Object.<string, string>). After that, we need one more function which accepts this list and create/update volume(s?) with these files.

configs/local/core/dashd.conf
strophy: Added a rule to accept RPC from the host machine

src/networks.js
strophy: OK, moved the version to the config schema

src/templates/renderServiceTemplates.js
strophy: I don't know, I thought it was good practice when fs operations are involved? Removed it now.

src/templates/renderServiceTemplates.js
strophy: OK, I tried to implement another function. I am 100% certain I have done the dependency injection wrong. I've tried for days, no matter what I do I just cannot get awilix to work. All of the code executes perfectly and the file is templated and written to disk, but I see errors like `renderServiceTemplates is not a function` or some Awilix injection error no matter what way I try it.

configs/local/core/dashd.conf
shumkov: Where? :)

strophy: https://github.com/dashevo/mn-bootstrap/pull/150/files#diff-617ba09d0095f6311d3e3ae245069bd8R22

templates/dashd.conf.template
```diff
@@ -0,0 +1,36 @@
...
+rpcpassword={{=it.rpcpassword}}
+rpcport={{=it.config.options.core.rpc.port}}
+
+rpcallowip=127.0.0.1:{{=it.config.options.core.rpc.port}}:{{=it.config.options.core.rpc.port}}
+rpcallowip=172.16.0.0/12
+rpcallowip=192.168.0.0/16
+
+rpcworkqueue=64
+rpcthreads=16
+
```
shumkov, line 28: Should we update it according to #153

templates/dashcore-node.json.template
```diff
@@ -0,0 +1,24 @@
...
+    "dashd",
+    "@dashevo/insight-api",
+    "web"
+  ],
+  "servicesConfig": {
+    "dashd": {
+      "connect": [{
+        "rpchost": "core",
+        "rpcport": {{=it.config.options.core.rpc.port}},
+        "rpcuser": "dashrpc",
```
shumkov, line 14: Let's move the user to config too?

src/templates/renderServiceTemplates.js
```diff
@@ -0,0 +1,28 @@
+const fs = require('fs');
+const dots = require('dot');
+const crypto = require('crypto');
+
```
shumkov, line 4: When you register a function using `asFunction` then it supposed to be a factory. Factory is a function that accepts dependencies and returns a function with logic. So this module should look like this:

```js
function renderServiceTemplatesFactory(writeServiceConfigs) {
    async function renderServiceTemplates(config, homeDirPath) {
       ...
    }

    return renderServiceTemplates;
}

module.exports = renderServiceTemplatesFactory;
```

You may find a lot of examples in this repo.

src/createDIContainer.js
```diff
@@ -63,6 +65,13 @@ async function createDIContainer() {
     // `configCollection` and `config` are registering on command init
   });
 
+  /**
+   * Templates
+   */
+  container.register({
```
shumkov, line 71: You need to register `writeServiceConfigs` too. There are no dependencies so I would use `asValue` helper to register it.

docker-compose.yml
```diff
@@ -6,14 +6,13 @@ services:
     restart: unless-stopped
     ports:
       - ${CORE_P2P_PORT:?err}:${CORE_P2P_PORT:?err} # P2P
+      - 127.0.0.1:${CORE_RPC_PORT:?err}:${CORE_RPC_PORT:?err} #RPC
     volumes:
-      - ./configs/${NETWORK:?err}/core/dashd.conf:/dash/.dashcore/dash.conf
+      - ~/.mn/${NETWORK:?err}/dashd.conf:/dash/.dashcore/dash.conf
```
shumkov, line 11: Shouldn't be a config name instead of network? Also, path to home dir is variable so we need to pass it here too.

src/templates/renderServiceTemplates.js
strophy: How did you learn this the first time you tried it? I was stuck for nearly two weeks on this :( and there is nothing in the [awilix docs](https://github.com/jeffijoe/awilix#asfunction) that explains that only factory functions can be used with `asFunction`... It is working now, but I still have no idea why it works.

src/createDIContainer.js
strophy: Why does it matter if a function has dependencies? It looks like `writeServiceConfigs` depends on `fs`, why should I use `asValue` there but not for `renderServiceTemplatesFactory`, which also depends only on external modules like `fs` and `dot`...?

docker-compose.yml
strophy: Set to config name for now. The home dir is currently not variable, but I can try to set it variable later.

templates/initiate_mongodb_replica.js.template
```diff
@@ -5,7 +5,7 @@ rs.initiate({
   members: [
     {
       _id: 0,
-      host: 'drive_mongodb:27017',
+      host: 'drive_mongodb:{{=it.platform.drive.mongodb.port}}',
```
shumkov, line 8: It's internal IP do we need to configure it?

src/config/Config.js
```diff
@@ -119,6 +119,7 @@ class Config {
   toEnvs() {
     return {
       COMPOSE_PROJECT_NAME: `dash_masternode_${this.getName()}`,
+      CONFIG: `${this.getName()}`,
```
shumkov, line 122: `CONFIG_NAME` ?

src/config/configJsonSchema.js
```diff
@@ -107,13 +189,47 @@ module.exports = {
           type: 'object',
           properties: {
             mongodb: {
-              $ref: '#/definitions/docker',
+              properties: {
+                docker: {
+                  $ref: '#/definitions/docker/properties/docker',
+                },
+                port: {
```
shumkov, line 196: Do we need to specify MongoDB port? It's just internal one

src/config/configJsonSchema.js
```diff
@@ -88,16 +116,70 @@ module.exports = {
...
+                  type: 'object',
+                  properties: {
+                    port: {
+                      $ref: '#/definitions/port',
+                    },
+                  },
+                  required: ['port'],
+                  additionalProperties: false,
+                },
+                txfilterstream: {
```
shumkov, line 158: txFilterStream is not part of api, it's a separate service actually.

src/config/configJsonSchema.js
```diff
@@ -107,13 +189,47 @@ module.exports = {
...
+                  $ref: '#/definitions/port',
+                },
+              },
+              required: ['port', 'docker'],
+              additionalProperties: false,
             },
             abci: {
               $ref: '#/definitions/docker',
             },
             tendermint: {
```
shumkov, line 206: tenderdash

docker-compose.platform.yml
```diff
@@ -6,7 +6,7 @@ services:
     restart: unless-stopped
     volumes:
       - drive_mongodb_data:/data/db
-      - ./docker/mongodb/initiate_mongodb_replica.js:/docker-entrypoint-initdb.d/initiate_mongodb_replica.js
+      - ~/.mn/${CONFIG:?err}/initiate_mongodb_replica.js:/docker-entrypoint-initdb.d/initiate_mongodb_replica.js
     extra_hosts:
       - "drive_mongodb:127.0.0.1"
     command: mongod --replSet driveDocumentIndices --bind_ip_all
```
shumkov, line 12: We need to pass MongoDB port

docker-compose.platform.yml
```diff
@@ -35,9 +35,10 @@ services:
     depends_on:
       - drive_abci
     ports:
-      - 26656:26656 # P2P
+      - ${PLATFORM_DRIVE_TENDERMINT_P2P_PORT:?err}:${PLATFORM_DRIVE_TENDERMINT_P2P_PORT:?err} # P2P
     volumes:
-      - ./configs/${NETWORK:?err}/drive/tendermint:/tendermint/config:ro
+      - ~/.mn/${CONFIG:?err}/genesis.json:/tendermint/config/genesis.json
```
shumkov, line 40: We need to pass home path

shumkov:
@strophy we also need to update ports in `startCore`.

docker-compose.yml
```diff
@@ -22,10 +21,10 @@ services:
...
     environment:
       - DEBUG=false
-      - RPCUSER=dashrpc
-      - RPCPASSWORD=password
+      - RPCUSER=${CORE_RPC_USER:?err}
+      - RPCPASSWORD=${CORE_RPC_PASSWORD:?err}
       - RPCHOST=core
-      - RPCPORT=20002
+      - RPCPORT=${CORE_RPC_PORT:?err}
       - NETWORK=${NETWORK?:err}
```
shumkov, line 28: Will it be `NETWORK_NAME`?

src/config/configJsonSchema.js
```diff
@@ -88,16 +116,70 @@ module.exports = {
...
             nginx: {
               $ref: '#/definitions/docker',
             },
             api: {
-              $ref: '#/definitions/docker',
+              properties: {
+                docker: {
+                  $ref: '#/definitions/docker/properties/docker',
+                },
+                jsonrpc: {
```
shumkov, line 138: jsonRpc

src/config/configJsonSchema.js
```diff
@@ -88,16 +116,70 @@ module.exports = {
...
+                  type: 'object',
+                  properties: {
+                    port: {
+                      $ref: '#/definitions/port',
+                    },
+                  },
+                  required: ['port'],
+                  additionalProperties: false,
+                },
+                grpc: {
```
shumkov, line 148: gRpc

src/core/createRpcClient.js
```diff
@@ -19,7 +19,6 @@ function createRpcClient(config = {}) {
     user: 'dashrpc',
     pass: 'password',
     host: '127.0.0.1',
-    port: 20002,
```
shumkov, line 22: it could be still `20002` by default

templates/dashd.conf.template
```diff
@@ -1,50 +1,57 @@
...
+{{?}}
+
+
+{{? it.network.name === 'local'}}
+regtest=1
+[regtest]
+{{?? it.network.name === 'evonet'}}
+devnet={{=it.network.name}}-{{=it.network.version}}
+[devnet]
 addnode=seed-1.evonet.networks.dash.org:20001
```
shumkov, line 50: They should be configurable as well.

templates/dashd.conf.template
```diff
@@ -1,50 +1,57 @@
...
+zmqpubrawtxlock=tcp://0.0.0.0:{{=it.core.zmq.port}}
+zmqpubhashblock=tcp://0.0.0.0:{{=it.core.zmq.port}}
+{{?}}
+
+
+{{? it.network.name === 'local'}}
+regtest=1
+[regtest]
+{{?? it.network.name === 'evonet'}}
+devnet={{=it.network.name}}-{{=it.network.version}}
```
shumkov, line 48: We shouldn't add `-{{=it.network.version}}` if `network.version` is null

src/listr/tasks/startNodeTaskFactory.js
```diff
@@ -46,6 +52,10 @@ function startNodeTaskFactory(dockerCompose) {
         enabled: () => isUpdate === true,
         task: async () => dockerCompose.pull(config.toEnvs()),
       },
+      {
+        title: 'Render service templates',
```
shumkov, line 56: I feel templates needs for other commands as well (e.g. `generateToAddress`), so probably we should build them in BaseCommand after we determine the current config.

src/createDIContainer.js
shumkov: If you use `asFunction` helper, awilix looks for dependencies in the passed function arguments. It calls a factory. The factory creates a function with injected dependencies. You passing not a factory but a normal function you need to use `asValue` helper which means "as is".

src/core/startCoreFactory.js
```diff
@@ -50,13 +51,19 @@ function startCoreFactory(
...
       ],
     );
 
-    const rpcClient = createRpcClient();
-
+    const rpcClient = createRpcClient(
+      {
+        port: config.get('core.rpc.port'), 
+        user: config.get('core.rpc.user'),
+        pass: config.get('core.rpc.password')
```
shumkov, line 63: Let's rename this option to `password`.

docker-compose.platform.yml
```diff
@@ -35,9 +35,10 @@ services:
     depends_on:
```
shumkov, line 35: let's rename service and volume to `drive_tenderdash`.

src/config/configJsonSchema.js
```diff
@@ -141,8 +257,19 @@ module.exports = {
...
+        name: {
+          type: 'string',
+          enum: Object.values(NETWORKS),
+        },
+        version: {
+          type: 'integer',
+          minimum: 0,
+        },
+      },
+      required: ['name'],
```
shumkov, line 271: We usually require all properties to keep config structure strict.

src/config/configJsonSchema.js
```diff
@@ -141,8 +257,19 @@ module.exports = {
...
-      type: 'string',
-      enum: Object.values(NETWORKS),
+      type: 'object',
+      properties: {
+        name: {
+          type: 'string',
+          enum: Object.values(NETWORKS),
+        },
+        version: {
+          type: 'integer',
```
shumkov, line 267: It's optional, so let's make it `type: ['string', 'null'],` as other optional params

docker-compose.platform.yml
strophy: I think that is done here?
https://github.com/dashevo/mn-bootstrap/pull/150/files/602e2bbb9cdba4e4457f9fa2c7cfa47f7fc3c8c8#diff-399795f6a9dce5cdec6c156808e7894827be79918aff4b7674b2249a550a49f9R8

src/listr/tasks/startNodeTaskFactory.js
strophy: This turned out to be the bug that had me stuck for hours trying to figure out why docker was writing dashd.conf as a dir owned by root (because it hadn't been rendered yet). Good catch!

src/core/startCoreFactory.js
strophy: I don't think we can do that, because [`dashd-rpc`](https://www.npmjs.com/package/@dashevo/dashd-rpc#usage) expects the `config` object to contain a key named `pass`. Or do you mean something else?

templates/dashd.conf.template
strophy: What do you mean by "they"?

src/core/startCoreFactory.js
shumkov: Got it. Let's keep it.

templates/dashd.conf.template
shumkov: Sorry, I meant `addnode` options should be configurable as well.

docker-compose.platform.yml
shumkov: Nope, there is how we connect to a replica node but not what port the node is listening on. We need to pass `--port` option to `mongod`.

strophy: I tried putting mongodb on a configurable port (e.g. 27027) and repeatedly ran into this error:

```
/usr/local/bin/docker-entrypoint.sh: running /docker-entrypoint-initdb.d/initiate_mongodb_replica.js
2020-10-14T03:04:31.192+0000 I  NETWORK  [listener] connection accepted from 127.0.0.1:33712 #2 (1 connection now open)
2020-10-14T03:04:31.192+0000 I  NETWORK  [conn2] received client metadata from 127.0.0.1:33712 conn2: { application: { name: "MongoDB Shell" }, driver: { name: "MongoDB Internal Client", version: "4.2.10" }, os: { type: "Linux", name: "Ubuntu", architecture: "x86_64", version: "18.04" } }
2020-10-14T03:04:31.220+0000 I  REPL     [conn2] replSetInitiate admin command received from client
2020-10-14T03:04:31.221+0000 E  REPL     [conn2] replSet initiate got NodeNotFound: No host described in new configuration 1 for replica set driveDocumentIndices maps to this node while validating { _id: "driveDocumentIndices", version: 1.0, members: [ { _id: 0.0, host: "drive_mongodb:27027" } ] }
2020-10-14T03:04:31.243+0000 I  NETWORK  [conn2] end connection 127.0.0.1:33712 (0 connections now open)
```

This made me realise that we don't actually need to add most of these ports to the config because they are internal to the named network, right? Since each network has its own namespace, they can use the same/default ports internally, only published ports need to be configurable so they don't conflict on the host. I think the list of published ports is:
- dashd P2P
- dashd RPC (host only, for status command)
- tendermint P2P
- nginx jsonrpc
- nginx grpc

Is my thinking right, can we remove a lot of this from config again?

shumkov: Yeah, you right.

strophy: OK, I cleaned up all config except relating to these 5 ports.

templates/dashd.conf.template
strophy: OK, added `seeds` and `sporkAddr` to config.

src/config/configJsonSchema.js
```diff
@@ -141,8 +201,27 @@ module.exports = {
...
+      type: 'object',
+      properties: {
+        name: {
+          type: 'string',
+          enum: Object.values(NETWORKS),
+        },
+        version: {
+          type: ['string', 'null'],
+        },
+        seeds: {
```
shumkov, line 213: I guess it should be in `core.p2p` section.

src/config/configJsonSchema.js
```diff
@@ -141,8 +201,27 @@ module.exports = {
...
+          type: 'array',
+          items: {
+            type: 'string'
```
shumkov, line 216: I would add a pattern to check `host:port`.

src/config/configJsonSchema.js
```diff
@@ -141,8 +201,27 @@ module.exports = {
...
+          },
+        },
+        sporkAddr: {
```
shumkov, line 219: Let's move it to `core.sporkAddress`

src/config/systemConfigs/systemConfigs.js
```diff
@@ -30,6 +35,12 @@ const baseConfig = {
         },
       },
       nginx: {
+        default: {
```
shumkov, line 38: http

templates/dashd.conf.template
```diff
@@ -1,55 +1,67 @@
...
+{{?}}
 
-sporkaddr=yQuAu9YAMt4yEiXBeDp3q5bKpo7jsC2eEj
+
+{{? it.network.name === 'local'}}
+regtest=1
+[regtest]
+{{?? it.network.name === 'evonet'}}
+devnet={{=it.network.name}}{{? it.network.version !== null}}-{{=it.network.version}}{{?}}
+{{? it.core.docker.image.includes('16') }}[devnet]{{?}}
```
shumkov, line 44: endsWith?

shumkov (commented):
We also need to put variables to tendermint configs.

templates/dashd.conf.template
strophy: OK, testing that it is not `15` for now. Unit testing is usually used to make sure code works properly, right? How does that work, do we use it anywhere?

docker-compose.yml
```diff
@@ -6,14 +6,13 @@ services:
...
       - ${CORE_P2P_PORT:?err}:${CORE_P2P_PORT:?err} # P2P
+      - 127.0.0.1:${CORE_RPC_PORT:?err}:${CORE_RPC_PORT:?err} #RPC
     volumes:
-      - ./configs/${NETWORK:?err}/core/dashd.conf:/dash/.dashcore/dash.conf
+      - ~/.mn/${CONFIG_NAME:?err}/dashd.conf:/dash/.dashcore/dash.conf
       - core_data:/dash
     command:
       - dashd
-      - -port=${CORE_P2P_PORT:?err}
       - -masternodeblsprivkey=${CORE_MASTERNODE_OPERATOR_PRIVATE_KEY}
```
shumkov, line 15: Should we set it in the dashd config too?

src/config/configJsonSchema.js
```diff
@@ -36,13 +39,47 @@ module.exports = {
...
-              minimum: 0,
+              $ref: '#/definitions/port',
+            },
+            seeds: {
+              type: 'array',
+              items: {
+                type: 'object',
+                properties: {
+                  host: {
+                    type: 'string',
```
shumkov, line 50: maxLength: 1

src/config/configJsonSchema.js
```diff
@@ -36,13 +39,47 @@ module.exports = {
...
           additionalProperties: false,
         },
+        rpc: {
+          type: 'object',
+          properties: {
+            port: {
+              $ref: '#/definitions/port',
+            },
+            user: {
+              type: 'string',
```
shumkov, line 71: maxLength: 1

src/config/configJsonSchema.js
```diff
@@ -141,8 +220,18 @@ module.exports = {
...
-      type: 'string',
-      enum: Object.values(NETWORKS),
+      type: 'object',
+      properties: {
+        name: {
+          type: 'string',
+          enum: Object.values(NETWORKS),
+        },
+        version: {
+          type: ['string', 'null'],
```
shumkov, line 230:
```
type: ['integer', 'null'],
minimum: 1
```

docker-compose.yml
```diff
@@ -6,14 +6,13 @@ services:
     restart: unless-stopped
     ports:
       - ${CORE_P2P_PORT:?err}:${CORE_P2P_PORT:?err} # P2P
+      - 127.0.0.1:${CORE_RPC_PORT:?err}:${CORE_RPC_PORT:?err} #RPC
```
shumkov, line 9: Do we want to publish the RPC port always?

src/core/startCoreFactory.js
```diff
@@ -35,6 +35,7 @@ function startCoreFactory(
       'dashd',
       '-conf=/dash/.dashcore/dash.conf',
       `-port=${config.get('core.p2p.port')}`,
```
shumkov, line 37: we have these options in dashd config I suppose

src/core/startCoreFactory.js
```diff
@@ -50,13 +51,19 @@ function startCoreFactory(
       'core',
       coreCommand,
       [
-        '--publish=20002:20002',
+        `--publish=${config.get('core.rpc.port')}:${config.get('core.rpc.port')}`,
```
shumkov, line 54: We already publish RPC port by default https://github.com/dashevo/mn-bootstrap/pull/150/files#r507904035

templates/dashd.conf.template
shumkov: Yeah, but in this case it's more like functional test when we are testing against a blackbox. We have unit and integration tests in almost all projects/repos. You can check Drive, DPP, or DAPI for example.

shumkov: I feel we could remove this legacy. We will update testnet/evonet to 0.16 soon so we don't need to support 15 version anymore.

docker-compose.yml
```diff
@@ -12,7 +12,6 @@ services:
       - core_data:/dash
     command:
```
shumkov, line 13: Let's just remove the command?

src/config/configJsonSchema.js
strophy: Why? The host string is usually more than 1 character.

src/config/configJsonSchema.js
strophy: Why? The RPC username is usually more than 1 character.

docker-compose.yml
strophy: Yes, I think so. It will be required for `mn status` commands. Advanced users can manually comment this out and start with docker-compose directly.

shumkov: Agree. Then we don’t need to pass aditional params in startCore

src/config/configJsonSchema.js
shumkov: Ops.. I meant minLength :)

src/config/configJsonSchema.js
shumkov: I meant minLength, sorry.

docker-compose.yml
strophy: Setting this in dashd config means functions that require the wallet like `registerMasternodeTask` and `generateToAddressTask` don't work because the masternode is still enabled in the conf file:

```
Error: You can not start a masternode with wallet enabled.
```

I'm trying to temporarily do `config.set('core.masternode.enable', false);` and set it back to its original state later, but I don't understand when the service templates are actually rendered. Is it possible to use a debugger to see the actual sequence in which the functions run? Is this what the call stack is? Does run() or or runWithDependencies() go first, or are they the same thing?

shumkov: `run` goes first and then calls `runWithDependencies`. It seems it's better to move it back to docker-compose.

src/core/startCoreFactory.js
strophy: Got stuck here for a long time, turns out `docker-compose run` doesn't automatically publish ports. I've replaced explicit port publishing with `--service-ports`. The RPC port is definitely required for the generate and register commands, and also for the status command (as described in the PR description).

src/core/startCoreFactory.js
strophy: Yes, these options have been removed here.

docker-compose.yml
strophy: This is necessary to work around the BLS key in the dash.conf file when wallet functions are used in core.

docker-compose.yml
strophy: OK, finally resolved this, see comments below on related topics.

docker-compose.platform.yml
```diff
@@ -114,12 +115,12 @@ services:
...
     ports:
-      - 3000:80 # JSON RPC and gRPC Web
-      - 3010:50051 # gRPC Native
+      - ${PLATFORM_DAPI_NGINX_HTTP_PORT:?err}:80 # JSON RPC and gRPC Web
+      - ${PLATFORM_DAPI_NGINX_GRPC_PORT:?err}:50051 # gRPC Native
     volumes:
-      - ./configs/${NETWORK:?err}/dapi/nginx/default.conf:/etc/nginx/conf.d/default.conf
-      - ./configs/${NETWORK:?err}/dapi/nginx/grpc.conf:/etc/nginx/conf.d/grpc.conf
-      - ./configs/${NETWORK:?err}/dapi/nginx/errors.grpc_conf:/etc/nginx/conf.d/errors.grpc_conf
+      - ~/.mn/${CONFIG_NAME:?err}/default.conf:/etc/nginx/conf.d/default.conf
```
shumkov, line 121: Let's export `MN_HOME_DIR` too?

src/config/configJsonSchema.js
```diff
@@ -36,13 +39,50 @@ module.exports = {
...
+            password: {
+              type: 'string',
+              minLength: 1,
             },
           },
-          required: ['port'],
+          required: ['port', 'user', 'password'],
           additionalProperties: false,
         },
+        sporkAddr: {
```
shumkov, line 83: Hate shortcuts. It seems we need `spork` object with `address` and `privateKey` properties.

templates/dashd.conf.template
```diff
@@ -35,23 +21,47 @@ listen=1
...
+{{~}}
+sporkaddr={{=it.core.sporkAddr}}
 
 # First 1000 blocks should be mined with lowest difficulty (like in regtest, so it's instant)
 # and first 500 blocks are mined with a block subsidity multiplied by 10
 # this allows to register MNs immediately (DIP3 activates on block 2)
 minimumdifficultyblocks=1000
 highsubsidyblocks=500
 highsubsidyfactor=10
+{{?? it.network.name === 'testnet'}}
```
shumkov, line 56: It must be before sections

templates/dashd.conf.template
```diff
@@ -35,23 +21,47 @@ listen=1
 dnsseed=0
 allowprivatenet=0
 
-printtoconsole=1
+{{? it.network.name === 'testnet'}}
```
shumkov, line 24: is there "not equal" operator?

templates/dashd.conf.template
```diff
@@ -35,23 +21,47 @@ listen=1
...
+{{?}}
 
-addnode=seed-1.evonet.networks.dash.org:20001
-addnode=seed-2.evonet.networks.dash.org:20001
-addnode=seed-3.evonet.networks.dash.org:20001
-addnode=seed-4.evonet.networks.dash.org:20001
-addnode=seed-5.evonet.networks.dash.org:20001
 
-sporkaddr=yQuAu9YAMt4yEiXBeDp3q5bKpo7jsC2eEj
+{{? it.network.name === 'local'}}
```
shumkov, line 39: We need a section for testnet

templates/dashd.conf.template
strophy: It's just plain JS, so the "not equal" operator is `!==`

templates/dashd.conf.template
strophy: Added

src/config/configJsonSchema.js
strophy: Added

docker-compose.platform.yml
strophy: How? I spent 3 hours fighting awilix to try and access the container to get this var, but it just doesn't work. Can you teach me how to use a debugger so I can see what the actual path is for vars like this? At the moment I am just guessing :(

docker-compose.platform.yml
```diff
@@ -35,9 +35,10 @@ services:
     depends_on:
       - drive_abci
     ports:
-      - 26656:26656 # P2P
+      - ${PLATFORM_DRIVE_TENDERMINT_P2P_PORT:?err}:26656 # P2P
```
shumkov, line 38: We need RPC port too (like in `master`):
- 127.0.0.1:${PLATFORM_DRIVE_TENDERMINT_RPC_PORT}:26657

src/config/Config.js
```diff
@@ -119,6 +119,8 @@ class Config {
   toEnvs() {
     return {
       COMPOSE_PROJECT_NAME: `dash_masternode_${this.getName()}`,
+      MN_HOME_DIR: '~/.mn', // this.container.resolve('homeDirPath'),  How do I get this variable?
```
shumkov, line 122: Maybe we should export it here? https://github.com/dashevo/mn-bootstrap/blob/6dd9a0b960fa7361bc0e4f03c6758a7082e32011/src/docker/DockerCompose.js#L266
And please don't forget about `config:envs` command.

src/core/startCoreFactory.js
```diff
@@ -50,13 +48,19 @@ function startCoreFactory(
       'core',
       coreCommand,
       [
-        '--publish=20002:20002',
+        '--service-ports',
```
shumkov, line 51: Do we need it since we already mapped it in `docker-compose.yml`?

strophy: Yes, because `docker-compose run` ignores the ports specified in the service configuration. See docs here: https://docs.docker.com/compose/reference/run/

shumkov: Oh.. I see. Good to know, thank you.

docker-compose.platform.yml
strophy: Added

src/config/Config.js
strophy: How and why do I export it there, what process did you use to find out this is the place where it needs to be exported? When is `getOptions(envs)` function called and how does it relate to `config:envs` and `Config.toEnvs()`?? How can I visualise these relationships between functions?

shumkov: Just pass home as a dependency to DocketCompose file so it will be available in getOptions. This method is called before envs passed to docker-compose cli. So it reasonable to enrich config envs with this system env in this place.

strophy: OK, I did what you said and it works, but I don't understand how. Also now nothing has changed in `Config.js`, and the variable doesn't appear in the output of `config:envs`. I still don't know what result you want to achieve with this, without knowing why we are doing it, I don't know what to do next.

shumkov: I'm sorry, I'm not so good at writing. Let me try one more time ). We do not have this homedir option in the config and it's kind of out of its scope logically. So, it makes sense just to add it somewhere on a lower level where we passing env variables to docker-compose CLI for example. There are two places where we are doing this:
- `DockerCompose#getOption` (bad naming for sure, must be something like `createOptions`) 
- `config:env` command

So we can enrich envs with this system one in these two places and logically it will be ok.

templates/dashd.conf.template
shumkov: So pity. It looks ugly a bit :(. Actually, you can use JS if conditions: https://github.com/olado/doT/blob/master/examples/advancedsnippet.txt#L41-L43

docker-compose.platform.yml
shumkov: To use debugger you should add `bin/mn` as you run script in run/debug settings and specify a needed command as arguments. I've attached a screenshot how it looks in WebStorm.
Screenshot 2020-11-03 at 20 13 33
I guess in VC must be about the same.

src/config/configJsonSchema.js
```diff
@@ -232,6 +232,10 @@ module.exports = {
                   required: ['port'],
                   additionalProperties: false,
                 },
+                genesisTime: {
```
shumkov, line 235: I would probably group all options from `genesis.json` into `genesis` object. Coz most of them actually changing in time, but we need to specify genesis one (how they were on block 1) to be able to sync.

src/config/configJsonSchema.js
```diff
@@ -232,6 +232,10 @@ module.exports = {
...
+                  type: ['string', 'null'],
+                  minLength: 1,
```
shumkov, line 237: I guess it's better to `format` keyword here.

src/config/configJsonSchema.js
```diff
@@ -232,6 +232,10 @@ module.exports = {
                   required: ['port'],
                   additionalProperties: false,
                 },
+                genesisTime: {
+                  type: ['string', 'null'],
```
shumkov, line 236: Actually, this one is always required.

src/config/configJsonSchema.js
```diff
@@ -272,7 +276,7 @@ module.exports = {
                   },
                 },
               },
-              required: ['docker', 'p2p', 'rpc', 'validators', 'persistentPeers'],
+              required: ['docker', 'p2p', 'rpc', 'genesisTime', 'validators', 'persistentPeers'],
```
shumkov, line 279: Let's move `persistentPeers` to `p2p`?

src/config/configJsonSchema.js
strophy: Then we have to set a `genesisTime` for `testnet`, even though the tendermint chain isn't running there yet?

src/config/configJsonSchema.js
strophy: You mean every option specified in the file? Or just the validators and chain_id? I don't know anything about tendermint, so I don't understand what happens when checking the genesis or what other options might be variable.

templates/dashd.conf.template
strophy: Not sure what bit looks ugly to you? You want me to change all my `{{? <condition>}} <output> {{?}}` syntax to `{{if ({{<condition>}}) { }} <output> {{ } }}` right? Isn't that a **lot** uglier?

docker-compose.platform.yml
strophy: Yes, it's exactly the same. I have no problems starting the debugger and using it. I just can't find the awilix container data, so I don't understand how to access it in different contexts.

src/config/Config.js
strophy: Hmm, maybe I was getting `DockerCompose#getOptions` confused with `Config#getOptions` while trying to use the stack trace in the debugger... I added it to `envs.js` instead of `Config.js` now, so it is added in the command and not in the class. Still not clear why we need to export this, since it is a constant and not a variable.

src/config/configJsonSchema.js
shumkov: Platform will be on testnet in a month and half :). Yeah, let’s put something here for now but we will be able to keep config consistent.

src/config/Config.js
shumkov: I remember we discussed making it an environment variable to allow people to specify the path. It makes sense for the deployment tool for example. We didn't do that yet?

templates/dashd.conf.template
shumkov: Actually I meant to change `{{? it.network.name === 'testnet'}}{{??}}output{{?}}` to `{{ if (it.network.name !== 'testnet') { }}output{{ } }}`

docker-compose.platform.yml
shumkov: Ah, I see. So it depends on how you register service in the container. If you use `asClass` then dependencies will be passed to `constructor` params which are the same as service names in the container. In case of services registered using `asFunction` then dependencies will be passed as function params. We have a special case with `runWithDependecies` in command classes. In `BaseCommand` we manually pass args, options, and dependencies to `runWithDependecies` https://github.com/dashevo/mn-bootstrap/blob/47e12a8b85fa51db3b1d1e326e8e9dab7cb5c5c4/src/oclif/command/BaseCommand.js#L99-L103

src/config/configJsonSchema.js
shumkov: Yeah, we need all of them in the config. So we can define the `genesis` property as an object that should accept the whole object from `genesis.json` file.

templates/dashd.conf.template
strophy: That syntax hurts my eyes! Can you explain what you mean in chat? My understanding is we need to output `[test]` if the network is set to testnet...

src/config/Config.js
strophy: No, I couldn't figure out how to pass command line flags to awilix. I thought we agreed to do it in a separate PR.

src/config/configJsonSchema.js
strophy: Should we validate the JSON with Ajv still, or treat it like a string? I can't get doT to interpolate an entire JSON value, it just returns `[object Object]`...

src/config/Config.js
shumkov: Ah I see. Actually you don't need to pass command line flags. All env variables in `proccess.env` so we usually pass  it to `createDIContainer` as an options param https://github.com/dashevo/js-drive/blob/v0.17-dev/scripts/abci.js#L10

src/config/configJsonSchema.js
shumkov: I think we don't need to validate it but just kept it there as an JSON object (not serialized to string). Just use `type: object` and do not define `properties`. May be in this case it's better to take this info and put result of `JSON.stringify` directly to the `genesys.json` file. So basically we don't need the template for this case.

templates/dashd.conf.template
shumkov: Yeah, we defo need. But I don't see it right now there. It's just empty `if` body and only `else` body has output. If you put `[testnet]` to `if` body I'm good with that. My point was that keeping an empty `if` body just to use `else` body is ugly.

strophy: `testnet=1` and `[test]` are in the `if` body - are we looking at the same code?

src/config/configJsonSchema.js
strophy: OK, found my syntax error with doT. We now treat it as an object and the template evaluates the entire object through JSON.stringify(), so it can be handled in the same way as the other templated files.

src/config/Config.js
strophy: I think I see, so `process.env` contains the current user environment and we can just check if a var is set there? I defined it to check if `MN_HOME_DIR` is set, is that right?

src/config/configJsonSchema.js
strophy: I think this is resolved now through different handling of the genesis data above?

src/config/Config.js
shumkov: Yeah, something like:

```js
const homeDirPath = process.env.MN_HOME_DIR || path.resolve(os.homedir(), '.mn');
```

templates/dashd.conf.template
shumkov: Hehe not sure. Just trust in your words :)))

src/config/configJsonSchema.js
shumkov: Great!

src/config/configJsonSchema.js
shumkov: Yep!

templates/config.toml.template
```diff
@@ -47,7 +47,7 @@ log_format = "plain"
 ##### additional base config options #####
 
 # Path to the JSON file containing the initial validator set and other meta data
-genesis_file = "data/genesis.json"
+genesis_file = "config/genesis.json"
```
shumkov, line 50: I guess it expected to be in `data` no?

strophy: Right! But the file will only be generated if it does not exist, and we can't mount it conditionally in docker-compose. Instead we will mount a blank dummy and use the mutable `data` dir to generate a new file if genesis data is blank.

strophy:
Automatic fetching the DPNS registration block works with `mn setup-for-local-development` now, but I am still getting correction errors syncing tendermint on evonet. External IP is set for tendermint now. Can you reproduce this?

strophy:
Do we need to set a new `devnet` network type in `networks.js`?

shumkov:
> Do we need to set a new `devnet` network type in `networks.js`?

Yep, let's add it.

shumkov:
> Can you reproduce this?
It works well for me.

shumkov (approved):
Good job! 🥇 Excited to have it done!

PR #119, opened 2020-08-11
feat: replace env files and presets with dynamic configs

strophy:
Adds a config command to persist config state across sessions. See #61 

## Issue being fixed or feature implemented
It is currently necessary to specify which network is being used in every single command. This feature stores the config state on disk and provides a number of interfaces with which to manage the configuration.

## What was done?
- [x] Described Config and ConfigCollection classes
- [x] Added ConfigRepostitory and read/write logic on command start/end
- [x] Added basic commands for managing configuration
- [x] Implemented config validation
- [x] Removed dotenv files and pass envs to Docker Compose directly
- [x] Data and config dirs depend on network name
- [x] Getting data from configs if options are missing and storing command results in config

## How Has This Been Tested?
Manually

## Breaking Changes
When complete, this will change the way each command is issued

## Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation

**For repository code-owners and collaborators only**
- [x] I have assigned this pull request to a milestone

Issue #68, opened 2020-06-10
Restart command


shumkov:
## Expected Behavior

Restart command could be useful in case of local preset with image build path options. So updating code you can restart the node and rebuild stuff.

## Current Behavior

You need make two commands: stop and then start.

## Possible Solution

Implement reset command.

## Alternatives Considered

None

## Additional Context

None

strophy:
To do this, I think we either need to store the config state somewhere (like in #61) or somehow use the `history` command to retrieve it?

shumkov:
Yeah, we need the `config` command

shumkov:
It's done on #152

PR #152, opened 2020-10-08
feat: restart command

strophy:
This PR implements the `restart` command described in #68 

Now that the config command was implemented in #119, the config state is stored on disk. This allows us to restart a node with a single command, running updated code or config on restart.

## What was done?
My approach here was just to combine the stop and start commands in a single new command. Oclif allows you to [call commands programmatically](https://oclif.io/docs/running_programmatically), but I'm not sure if this is a good idea here, given we would just be passing the same arguments around anyway. 

## How Has This Been Tested?
- [x] Tested restarting nodes with different confs
- [x] Tested restarting after a config change was made to the running node

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ Distribution package for Dash Masternode installation
   - [Configure node](#configure-node)
   - [Start node](#start-node)
   - [Stop node](#stop-node)
+  - [Restart node](#restard-node)
   - [Register masternode](#register-masternode)
   - [Reset data](#reset-data)
   - [Development](#development)
@@ -119,6 +120,27 @@ OPTIONS
 To stop a node:
 ```bash
 $ mn stop
+```
+
+### Restart node
+
+The `restart` command is used to restart a node with the default or specified config.
+
+```
+USAGE
+  $ mn restart
+
+OPTIONS
+  -f, --full-node                                  start as full node
+  -u, --update                                     download updated services before start
+  --config=config                                  configuration name to use
+  --dapi-image-build-path=dapi-image-build-path    dapi's docker image build path
+  --drive-image-build-path=drive-image-build-path  drive's docker image build path
+```
+
+To update services and restart a masternode:
+```bash
+$ mn restart -u
 ```
 
 ### Register masternode
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ Distribution package for Dash Masternode installation
   - [Restart node](#restard-node)
   - [Register masternode](#register-masternode)
   - [Reset data](#reset-data)
+  - [Full node](#full-node)
   - [Development](#development)
   - [Docker Compose](#docker-compose)
 - [Contributing](#contributing)
--- a/README.md
+++ b/README.md
@@ -218,6 +218,14 @@ To show the host status:
 $ mn status:host
 ```
 
+### Full node
+
+It is also possible to start a full node instead of a masternode. Modify the config setting as follows:
+
+```bash
+mn config:set core.masternode.enable false
+```
+
 ### Development
 
 When developing on a standalone node (a config specifying the `local` network), `setup-for-local-development` can be used to generate some dash, register a masternode and populate the node with the data required for local development.
--- a/README.md
+++ b/README.md
@@ -89,7 +89,6 @@ The `start` command is used to start a node with the default or specified config
 USAGE
   $ mn start
 OPTIONS
-  -f, --full-node                                  start as full node
   -u, --update                                     download updated services before start
   --config=config                                  configuration name to use
   --dapi-image-build-path=dapi-image-build-path    dapi's docker image build path
@@ -99,11 +98,6 @@ OPTIONS
 To start a masternode:
 ```bash
 $ mn start
-```
-
-To start a full node:
-```bash
-$ mn start -f
 ```
 
 ### Stop node
@@ -131,7 +125,6 @@ USAGE
   $ mn restart
 
 OPTIONS
-  -f, --full-node                                  start as full node
   -u, --update                                     download updated services before start
   --config=config                                  configuration name to use
   --dapi-image-build-path=dapi-image-build-path    dapi's docker image build path
--- a/src/commands/restart.js
+++ b/src/commands/restart.js
@@ -1,0 +1,82 @@
+const { Listr } = require('listr2');
+
+const { flags: flagTypes } = require('@oclif/command');
+
+const BaseCommand = require('../oclif/command/BaseCommand');
+
+const MuteOneLineError = require('../oclif/errors/MuteOneLineError');
+
+class RestartCommand extends BaseCommand {
+  /**
+   * @param {Object} args
+   * @param {Object} flags
+   * @param {DockerCompose} dockerCompose
+   * @param {startNodeTask} startNodeTask
+   * @param {Config} config
+   * @return {Promise<void>}
+   */
+  async runWithDependencies(
+    args,
+    {
+      'full-node': isFullNode,
+      update: isUpdate,
+      'drive-image-build-path': driveImageBuildPath,
+      'dapi-image-build-path': dapiImageBuildPath,
+    },
+    dockerCompose,
+    startNodeTask,
+    config,
+  ) {
+    const isMinerEnabled = config.get('core.miner.enable');
+
+    const tasks = new Listr(
+      [
+        {
+          title: 'Stop node',
+          task: async () => dockerCompose.stop(config.toEnvs()),
+        },
+        {
+          title: `Start ${isFullNode ? 'full node' : 'masternode'}`,
+          task: () => startNodeTask(
+            config,
+            {
+              isFullNode,
+              driveImageBuildPath,
+              dapiImageBuildPath,
+              isUpdate,
+              isMinerEnabled,
+            },
+          ),
+        },
+      ],
+      {
+        rendererOptions: {
+          clearOutput: false,
+          collapse: false,
+          showSubtasks: true,
+        },
+      },
+    );
+
+    try {
+      await tasks.run();
+    } catch (e) {
+      throw new MuteOneLineError(e);
+    }
+  }
+}
+
+RestartCommand.description = `Restart masternode
+...
+Restart masternode with specific preset
+`;
+
+RestartCommand.flags = {
+  ...BaseCommand.flags,
+  'full-node': flagTypes.boolean({ char: 'f', description: 'start as full node', default: false }),
+  update: flagTypes.boolean({ char: 'u', description: 'download updated services before start', default: false }),
+  'drive-image-build-path': flagTypes.string({ description: 'drive\'s docker image build path', default: null }),
+  'dapi-image-build-path': flagTypes.string({ description: 'dapi\'s docker image build path', default: null }),
+};
+
+module.exports = RestartCommand;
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -121,6 +122,27 @@ To stop a node:
...
+### Restart node
+
+The `restart` command is used to restart a node with the default or specified config.
+
+```
+USAGE
+  $ mn restart
+
+OPTIONS
+  -f, --full-node                                  start as full node
```
shumkov, line 134: This kinda strange to see such a flag in restart command. Perhaps, we should move it to config?

strophy: ok, moved to config

src/listr/tasks/startNodeTaskFactory.js
```diff
@@ -26,7 +26,7 @@ function startNodeTaskFactory(dockerCompose) {
   function startNodeTask(
     config,
     {
-      isFullNode,
+      isMasternode,
```
shumkov, line 29: We don’t need to pass it as an argument since we can read it from config

src/commands/restart.js
```diff
@@ -0,0 +1,81 @@
...
+    dockerCompose,
+    startNodeTask,
+    config,
+  ) {
+    const isMinerEnabled = config.get('core.miner.enable');
+    const isMasternode = config.get('core.masternode.enable');
+
+    const tasks = new Listr(
+      [
+        {
```
shumkov, line 34: I would check that node is running otherwise the restart command will start the node even if it’s stopped.

README.md
```diff
@@ -100,11 +100,6 @@ To start a masternode:
 $ mn start
 ```
 
-To start a full node:
```
shumkov, line 103: I guess it still worth documenting this. We could add a separate section for that after ”Register mssternode”

src/listr/tasks/startNodeTaskFactory.js
strophy: I guess the same applies to `isMinerEnabled` then right? Both miner and masternode are now read from config instead of passed as arguments.

README.md
strophy: Added a bit further below, since this is not technically a command

src/commands/restart.js
strophy: Isn't this the behaviour we want? In my mind this function is used after fixing something in the config that is causing the node to run incorrectly, e.g. in stopped or restarting state. If we do want to check it is running, which service should we check? All of them?

shumkov: Hm... I got used if you try to restart stopped thing you get an error that the service is not started. Is not it? You can use `isServiceRunning` without arguments. But probably, it’s better to put in `startNodeTask`.

src/commands/setup-for-local-development.js
```diff
@@ -53,6 +53,10 @@ class SetupForLocalDevelopmentCommand extends BaseCommand {
               title: 'Register masternode',
               task: () => registerMasternodeTask(config),
             },
+            {
```
shumkov, line 56: Not sure we need the miner here.

src/listr/tasks/startNodeTaskFactory.js
shumkov: Defo! Nice catch!

src/commands/restart.js
```diff
@@ -0,0 +1,78 @@
...
+   * @param {DockerCompose} dockerCompose
+   * @param {startNodeTask} startNodeTask
+   * @param {Config} config
+   * @return {Promise<void>}
+   */
+  async runWithDependencies(
+    args,
+    {
+      update: isUpdate,
+      'drive-image-build-path': driveImageBuildPath,
```
shumkov, line 22: I would move these two options to the config as well in a separate PR later. What do you think?

src/listr/tasks/startNodeTaskFactory.js
```diff
@@ -26,15 +26,15 @@ function startNodeTaskFactory(dockerCompose) {
...
       driveImageBuildPath = undefined,
       dapiImageBuildPath = undefined,
       isUpdate = undefined,
-      isMinerEnabled = undefined,
     },
   ) {
     // Check external IP is set
     config.get('externalIp', true);
+    const isMinerEnabled = config.get('core.miner.enable');
+    const isMasternode = config.get('core.masternode.enable');
```
shumkov, line 37: I would just move it a line above of using it (52)

src/listr/tasks/startNodeTaskFactory.js
```diff
@@ -26,15 +26,15 @@ function startNodeTaskFactory(dockerCompose) {
...
-      isFullNode,
       driveImageBuildPath = undefined,
       dapiImageBuildPath = undefined,
       isUpdate = undefined,
-      isMinerEnabled = undefined,
     },
   ) {
     // Check external IP is set
     config.get('externalIp', true);
+    const isMinerEnabled = config.get('core.miner.enable');
```
shumkov, line 36: Let's add an extra line between, just to increase readability.

src/commands/setup-for-local-development.js
strophy: I added this because on line 64 I had to remove `isMinerEnabled: true` so it works with the new arguments taken by `startNodeTask`

src/commands/restart.js
strophy: I don't understand your logic here. If we check the services are stopped in `startNodeTask` then in the `restart` command the stop node task will have already completed, so it will always evaluate `true`.

In general, during testing I often run into cases where a service is restarting because of a config problem. All our services are configured with `restart: unless-stopped` in docker-compose, so the only possible states are `running` or `restarting`. If I was to run `mn restart` in this case it would fail if we check for started services first, because the service is always restarting. Which means I am back to typing `mn stop` then `mn start` again anyway. Right?

strophy: I added a step in `startNodeTask` to check for stopped services, but the `this.error` syntax does not work here, including for the other error checks in this function. It returns `this.error is not a function`. I still don't understand how errors should be handled with all the dependencies, so I don't know how to fix this.

src/commands/restart.js
strophy: Yes, sure. But update should remain a CLI argument, right?

shumkov: Right.

src/commands/setup-for-local-development.js
shumkov: Yeah, but we don't need a miner for this command I believe :)

shumkov: Oh right :(... But shouldn't update config just to run this exact task. It seems we still need this option for `startNodeTask` just to change behavior for this specific call. So it should be like this there:

```js
if (isMinerEnabled === undefined) {
    isMinerEnabled = config.get('core.miner.enable');
}
```

src/commands/restart.js
shumkov: Yeah, if service is running (containers are present) then you should be able to use restart.

You just calling `this.error` in wrong context. It's a method of Oclif command, not `Listr`. For Listr task you just need to throw an error (`throw new Error('Node is not started')`);

src/listr/tasks/startNodeTaskFactory.js
```diff
@@ -46,10 +46,19 @@ function startNodeTaskFactory(dockerCompose) {
         enabled: () => isUpdate === true,
         task: async () => dockerCompose.pull(config.toEnvs()),
       },
+      {
+        title: 'Check for running services',
```
shumkov, line 50: "Check node is not started" or something like this.

src/commands/setup-for-local-development.js
strophy: `registerMasternodeTask` already updates the config with the BLS key, so I think it doesn't matter if we also change the miner setting? And miner is defined as `false` in the system config, so we can't check for `undefined`...

shumkov: The register command supposed obtain and store/output key, but here it will be a side effect. I don't need the miner, then run this command, and the miner is enabled for some reason. It’s strange.

strophy: Oh I understand now, the miner is needed while the `setup-for-local-development` task is running, but should not be in the config once setup is done, right? Changes done as you explained.

src/listr/tasks/startNodeTaskFactory.js
```diff
@@ -58,7 +70,7 @@ function startNodeTaskFactory(dockerCompose) {
 
           if (driveImageBuildPath || dapiImageBuildPath) {
             if (config.get('network') === NETWORKS.TESTNET) {
-              this.error('You can\'t use drive-image-build-path and dapi-image-build-path options with testnet network', { exit: true });
+              throw new Error('You can\'t use drive-image-build-path and dapi-image-build-path options with testnet network', { exit: true });
```
shumkov, line 73: `Error` constructor accepts `filename` as the second param. You don't need to pass anything except `message` (the first argument)

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error

src/listr/tasks/startNodeTaskFactory.js
```diff
@@ -26,18 +26,18 @@ function startNodeTaskFactory(dockerCompose) {
...
-      isMinerEnabled = undefined,
     },
   ) {
     // Check external IP is set
     config.get('externalIp', true);
+    
+    const isMinerEnabled = config.get('core.miner.enable');
 
     if (isMinerEnabled === true && config.get('network') !== NETWORKS.LOCAL) {
-      this.error(`'core.miner.enabled' option supposed to work only with local network. Your network is ${config.get('network')}`, { exit: true });
```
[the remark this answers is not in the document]
strophy: I changed it because it wasn't working, did you test it? Here is my output on testnet with miner enabled:

```
# with this.error()
strophy@X250:~/Code/mn-bootstrap$ mn start
✖ this.error is not a function 

# with throw new Error()
strophy@X250:~/Code/mn-bootstrap$ mn start
✖ 'core.miner.enabled' option only works with local network. Your network is testnet. 
```

How do I find out what the current context is? It's impossible to `console.log(this)` because of listr overwriting any console output, and the debugger shows `Error` but not `error` as a function of `this`, so I still don't know what `this` actually refers to.

src/listr/tasks/startNodeTaskFactory.js
strophy: Fixed, is there a way to see the parameters required by a function in the IDE? I can't get this working in VS Code because using awilix seems to break the "Go to definition" function for all injected functions.

src/listr/tasks/startNodeTaskFactory.js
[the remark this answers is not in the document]
shumkov: Oh... my bad. This is the context of listr task too since it's in `startNodeTask` what we pass to Listr as a task :).

You can understand the context usually just by reading the code. If it's not clear you can run the command with debugger. In WebStorm it's easy. Just add a Node.JS run configuration with `bin/mn` as a start file and pass `reset` as application params.

shumkov (approved):
Looks good! 👍

src/listr/tasks/startNodeTaskFactory.js
shumkov: You can use "go to function" feature by the function name.

PR #199, opened 2020-12-08
feat: config versions

strophy:
Store version of package.json used to create a config, and check that version against current package.json version during config validation.

This means it will no longer be necessary to drop config and start again when new versions are released.

## What was done?
Added a version to the config file schema and constructor, and check that version against the current package.json version to allow for transformations (yet to be defined) to the config as new settings are added.

## How Has This Been Tested?
Tested in debugger and by looking at the generated file. Validation works and the conditional code is triggered.

## Checklist:


src/config/ConfigCollection.js
```diff
@@ -53,6 +54,27 @@ class ConfigCollection {
...
+
+    return this;
+  }
+
+  /**
+   * Get current config version if set
+   *
+   * @returns {string|null}
+   */
+  getConfigVersion() {
```
shumkov, line 74: Config version could be confusing. Like the previous version represents previous changes? ConfigFormatVersion?

src/config/configFile/ConfigJsonFileRepository.js
```diff
@@ -46,6 +48,10 @@ class ConfigJsonFileRepository {
       throw new InvalidConfigFileFormatError(this.configFilePath, error);
     }
 
+    if (packageJson.version !== configFileData.configVersion) {
+      // do config upgrade
```
shumkov, line 52: Yeah, we need to run migration instructions here.. You can use the same format as we saw in that config library.

src/config/configFile/ConfigJsonFileRepository.js
```diff
@@ -46,6 +48,10 @@ class ConfigJsonFileRepository {
...
+    }
+
```
shumkov, line 54: Don't forget to pass the actual version down below.

src/config/ConfigCollection.js
strophy: Using `configFormatVersion`

src/config/configFile/ConfigJsonFileRepository.js
strophy: I agree, I really like the way migrations are handled here: https://github.com/sindresorhus/conf#migrations

I'm trying to learn from recent work scope mistakes and make a plan before beginning work ;) Does this look like the right plan for replacing our config system with `sindresorhus/conf`?

1. Remove Config, ConfigCollection and ConfigJsonFileRepository classes and replace with Conf
2. Either refactor function calls or write a wrapper for Conf with the same function calls we current support
3. Provide existing configFileJsonShema (with explicit link to configJsonSchema) to Conf as schema
4. Delete all other error functions and related code
5. Should we remove the config command or write functions that map to the Conf API? The `conf-cli` oclif plugin is also an option to access Conf from the CLI, but it works a bit different to our current system:
```
strophy@X250:~/Code/mn-bootstrap$ mn conf --help
manage configuration

USAGE
  $ mn conf [KEY] [VALUE]

ARGUMENTS
  KEY    key of the config
  VALUE  value of the config

OPTIONS
  -d, --cwd=cwd          config file location
  -d, --delete           delete?
  -h, --help             show CLI help
  -k, --key=key          key of the config
  -n, --name=name        config file name
  -p, --project=project  project name
  -v, --value=value      value of the config
```

It does most things in one command and doesn't require subcommands like get/set/list, and uses flags for things like delete. I'm not sure how we would handle commands like `config:reset`.

strophy: I'm currently stuck due to this issue: https://github.com/sindresorhus/conf/issues/80

Is there some way we can work around this?

shumkov: From my perspective, it's just easier to copy migration functionality instead of replacing the whole thing since we already implemented it :)

strophy: From your perspective, I suppose it is small and simple ;) But from my perspective, I can't understand how Sindre manages migrations in Conf, somehow he has access to setters that do not require a valid config to run. Also I think his code is written in TypeScript with no comments, so I can't really understand it. I think our design cannot produce a valid config file without having valid configs first, but to run the upgrade I think we need to modify the raw JSON read from the file, split it into configs and pass these to the Config class for validation? But when do we copy new variables, and where do we get the values? From the upgraded systemConfigs object? What if one of the configs has a name that doesn't match a system config?

I already spent almost 3 full days trying to get Conf working and made some progress... I thought you told me [here](https://github.com/dashevo/mn-bootstrap/pull/199#discussion_r539071730) to use the Conf format instead of just copying the migrate function? I don't understand how to do this the way you want it done at the moment. I really need a bit more guidance to get started on the right approach...

src/config/migrateConfigOptions.js
```diff
@@ -0,0 +1,19 @@
...
+const configOptionMigrations = require('./configOptionMigrations');
+
+function migrateConfigOptions(name, options, fromVersion, toVersion) {
+  if (fromVersion === toVersion) {
+    return options;
+  }
+
+  return Object.keys(configOptionMigrations)
+    .filter((version) => (semver.gt(version, fromVersion) && semver.lte(version, toVersion)))
+    .sort()
```
shumkov, line 12: we need to pass semver.compare function here.

strophy: I still don't understand how I can inspect the object/array when it is being processed in sequence like this. The output of each step in the function is not visible in the debugger, all I can see is the initial object passed to `Object.keys`. I can't see the how the object is transformed after each step. What is the right way to do this?
image

shumkov: I usually check what arguments I'm getting in each function so you can figure out what they return

strophy: Thanks, I finally understand how this works now :+1:

src/config/migrateConfigOptions.js
```diff
@@ -0,0 +1,19 @@
...
+const configOptionMigrations = require('./configOptionMigrations');
+
+function migrateConfigOptions(name, options, fromVersion, toVersion) {
+  if (fromVersion === toVersion) {
+    return options;
+  }
+
+  return Object.keys(configOptionMigrations)
+    .filter((version) => (semver.gt(version, fromVersion) && semver.lte(version, toVersion)))
+    .sort((v1, v2) => semver.compare(v1, v2))
```
shumkov, line 12:
```suggestion
    .sort(semver.compare)
```

shumkov (approved):
Good job! 👍

PR #200, opened 2020-12-09
feat: initialize Tenderdash

shumkov:

Introduce Tenderdash with initialization step. For different environments, different tenderdash configs could be preset so we need to generate dynamically missed once. 

## What was done?

- Move Node Key and Validator Key to config since they not always must be dynamically created
- Generate and fulfill Tenderdash configs if it's empty
- Used `dashpay/tenderdash` image
- Pass initial core chain locked height to Drive
- Some minor fixes

## How Has This Been Tested?

Running setup-for-local-development with https://github.com/dashevo/js-drive/pull/428

## Breaking Changes

None

## Checklist:


docker-compose.platform.yml
```diff
@@ -32,21 +32,21 @@ services:
...
     ports:
-      - ${PLATFORM_DRIVE_TENDERMINT_P2P_PORT:?err}:26656 # P2P
-      - 127.0.0.1:${PLATFORM_DRIVE_TENDERMINT_RPC_PORT:?err}:26657 # RPC
+      - ${PLATFORM_DRIVE_TENDERDASH_P2P_PORT:?err}:26656 # P2P
+      - 127.0.0.1:${PLATFORM_DRIVE_TENDERDASH_RPC_PORT:?err}:26657 # RPC
     volumes:
-      - ${MN_HOME_DIR:?err}/${CONFIG_NAME:?err}/genesis.json:/tendermint/config/genesis.json
-      - ${MN_HOME_DIR:?err}/${CONFIG_NAME:?err}/config.toml:/tendermint/config/config.toml
-      - drive_tendermint_homedir:/tendermint
-    entrypoint: sh -c "/usr/bin/tendermint init && /usr/bin/tendermint node"
```
antouhou, line 49: Don't we need an `entrypoint` anymore? Is it a default one now?

shumkov: Yeah, no idea why we didn't use it before :)

antouhou (dismissed):
👍

docker-compose.platform.yml
shumkov: Oh! Actually, we need it. We need to force init even if the config folder is already created with mounted configs.

shuplenkov (approved):
Looks legit!

docker-compose.platform.yml
```diff
@@ -32,21 +32,20 @@ services:
...
+      - ${MN_HOME_DIR:?err}/${CONFIG_NAME:?err}/tenderdash:/tenderdash/config
```
strophy, line 47: config dir should be mounted read-only?

src/tenderdash/initializeTenderdashNodeFactory.js
```diff
@@ -0,0 +1,113 @@
...
+      ].join('&&');
+
+      const [result] = await docker.run(
+        'alpine:latest',
+        ['sh', '-c', command],
+        writableStream,
+        {
+          HostConfig: {
+            AutoRemove: true,
+            Binds: [`${volumeNameFullName}:/tenderdash/data`],
```
strophy, line 60: I think the problem you are trying to fix here is `/tendermint` is being created as root? It shouldn't be necessary to `chown` if you mount the volume at the mountpoint specified in the container, see discussion [here](https://github.com/dashevo/mn-bootstrap/pull/123#discussion_r478167920). Mounting at a deeper mountpoint will create parent dirs in the volume as root in your alpine container. Since the tenderdash container specifies its own volume already, the result looks like this (note the last volume):
```
"Mounts": [
            {
                "Type": "bind",
                "Source": "/home/strophy/.mn/evonet/tenderdash",
                "Destination": "/tenderdash/config",
                "Mode": "rw",
                "RW": true,
                "Propagation": "rprivate"
            },
            {
                "Type": "volume",
                "Name": "dash_masternode_evonet_drive_tenderdash_data",
                "Source": "/var/lib/docker/volumes/dash_masternode_evonet_drive_tenderdash_data/_data",
                "Destination": "/tenderdash/data",
                "Driver": "local",
                "Mode": "rw",
                "RW": true,
                "Propagation": ""
            },
            {
                "Type": "volume",
                "Name": "ee14dce8ede3fa522ca954433681ca85866a319121b3abc58a5ce03c28d21b8d",
                "Source": "/var/lib/docker/volumes/ee14dce8ede3fa522ca954433681ca85866a319121b3abc58a5ce03c28d21b8d/_data",
                "Destination": "/tenderdash",
                "Driver": "local",
                "Mode": "",
                "RW": true,
                "Propagation": ""
            }
        ],
```

To avoid this hack, we could mount at `/tenderdash` instead, and either: 

- delete the `/tenderdash/config` files later after they are copied to the config object
- put another mount to catch the generated data in `/tenderdash/config` and drop it later

strophy (commented):
With local config, `mn setup-for-local-development` fails for me [here](https://github.com/dashevo/mn-bootstrap/blob/fc1b15e012f86b3d8197f4a708eb7567661caac9/src/listr/tasks/platform/initTaskFactory.js#L70) with: 
`Failed to perform standard injections with reason: Plugin ChainPlugin of type Standard onStart failed: No available addresses for retry: 13 INTERNAL: Error: socket hang up at new RequestError (/node_modules/request-promise-core/lib/errors.js:14:15)`

With evonet config, `mn start` fails to start the tenderdash container with the error: `ERROR: failed to create node: error reading GenesisDoc at /tenderdash/config/genesis.json: genesis doc must include initial_core_chain_locked_height greater than 0 (got 0)`

I think this is because `priv_validator_state.json` contains the following:
```
{
  "height": "0",
  "round": 0,
  "step": 0
}
```

src/tenderdash/initializeTenderdashNodeFactory.js
```diff
@@ -0,0 +1,113 @@
...
+
+    const { COMPOSE_PROJECT_NAME: composeProjectName } = config.toEnvs();
+    const volumeName = 'drive_tenderdash_data';
+    const volumeNameFullName = `${composeProjectName}_${volumeName}`;
+
+    const volume = docker.getVolume(volumeNameFullName);
+
+    const isVolumeDefined = await volume.inspect()
+      .then(() => true)
+      .catch(() => false);
```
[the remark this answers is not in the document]
shumkov: This is an exception 😅

docker-compose.platform.yml
shumkov: I guess so. When we start docker compose it must be initialized already

src/tenderdash/initializeTenderdashNodeFactory.js
shumkov: Actually, the problem is that volume FS is under root, but we creating tenderdash home dir as `tmuser`. That's why we need to set the correct user for tenderdash volume first.

strophy: I don't think this is true, we had exactly this problem with the `.dashcore` directory in the core image earlier this year. The tenderdash home dir already [exists](https://github.com/dashevo/tenderdash/blob/master/DOCKER/Dockerfile#L33) in the image with the correct owner. You are manually creating it again in a new volume with the wrong owner and then changing the owner, which is not necessary imo. If you mount it at the `/tenderdash` mountpoint, it will take the permissions already specified in the image. I have proven this locally as described above with a bind mount, but I'm getting an error "1" with no content from the writableStream when trying to implement it with a temp volume instead, and I don't know how to debug it...

I guess we can stick with your approach for now and create an issue to see if I can figure out a different way later?

jawid-h (approved):
Seems legit

shuplenkov (approved):
Looks good!

src/tenderdash/initializeTenderdashNodeFactory.js
shumkov: Fixed!

PR #249, opened 2021-01-20
feat: hard and soft resets, --platform-only option

strophy:
This PR adds more functionality to the reset command and renames the `config:reset` command to `reset --hard`.

Resetting configuration required running `setup` again after reset to regenerate platform init data. This is bad UX, because the user only expects to use `setup` to enter initial configuration. As a result, `reset` was implemented to leave platform in an initialized state, while `reset --hard` drops all containers and config as well. The `config:reset` command was removed.

## What was done?
- Remove `config:resest`
- Added platform init step to reset command
- Add `--hard` argument to also reset config
- Add `--platform-only` argument to only process non-core containers
- Refactor and rename commands as necessary

## How Has This Been Tested?
- Tested against testnet config
- Tested with both testnet and evonet config nodes running

## Breaking Changes
`config:reset` command no longer exists

## Checklist:


strophy:
Note that the `reset` command implements `docker-compose down` while the `reset --platform-only` command uses `docker-compose rm -f -v -s`. But `docker-compose rm -v` documentation states that this option "Remove any anonymous volumes attached to containers" - it will not remove named volumes. This is ok if the config stays the same, but if you do `reset --hard --platform-only` it results in errors because the dirty volumes are still there. Should we add a step to clean up volumes? Or should we be using anonymous volumes here?

shumkov:
@strophy Yes, we need to drop named volumes as well for hard and soft. But in the case of soft, we need to init tenderdash after that.

src/listr/tasks/tenderdash/initTaskFactory.js
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * @param {Config} config
+   * @return {Listr}
+   */
+  function tenderdashInitTask(
+    config,
+  ) {
+    return new Listr([
+      {
+        title: 'Validate keys and volumes',
```
shumkov, line 21: Generate node keys and data

src/listr/tasks/tenderdash/initTaskFactory.js
```diff
@@ -0,0 +1,58 @@
...
+        task: async () => {
+          const isValidatorKeyEmpty = Object.keys(config.get('platform.drive.tenderdash.validatorKey')).length === 0;
+          const isNodeKeyEmpty = Object.keys(config.get('platform.drive.tenderdash.nodeKey')).length === 0;
+          const isGenesisEmpty = Object.keys(config.get('platform.drive.tenderdash.genesis')).length === 0;
+
+          const { Volumes: existingVolumes } = await docker.listVolumes();
+          const { COMPOSE_PROJECT_NAME: composeProjectName } = config.toEnvs();
+          const isDataVolumeMissing = !existingVolumes.find((v) => v.Name === `${composeProjectName}_drive_tenderdash`);
+
+          if (isValidatorKeyEmpty || isNodeKeyEmpty || isNodeKeyEmpty || isDataVolumeMissing) {
```
shumkov, line 31: You can do `task.skip` and `return` if you meet the opposite condition.

strophy: OK, I think I inverted the logic correctly and everything works. I'm not sure where I am supposed to put `return`, it just creates a lot of errors about "Unnecessary return statement." or "Expected to return a value at the end of async arrow function 'task'."

strophy:
Thanks for the comments :) I implemented the final logic for `mn reset --platform-only --hard`, did a lot of refactoring and re-tested everything, let me know what you think.

src/commands/reset.js
```diff
@@ -8,30 +10,76 @@ class ResetCommand extends BaseCommand {
...
+        task: async () => dockerCompose.stop(config.toEnvs()),
+      },
+      {
+        title: 'Reset platform data',
+        enabled: () => isPlatformOnlyReset,
+        task: () => (
+          new Listr([
+            {
+              title: 'Remove platform services and associated data',
+              task: async () => dockerCompose.rmPlatformOnly(config.toEnvs()),
```
shumkov, line 47: DockerCompose is an infrastructure class that shouldn't know about specific business logic. I would recommend moving this function out.

src/commands/reset.js
```diff
@@ -8,30 +10,76 @@ class ResetCommand extends BaseCommand {
...
+            },
+            {
+              title: 'Clean up platform volumes',
+              task: async () => docker.pruneVolumes(),
```
shumkov, line 51: I guess `down -v` must remove volumes as well

src/commands/reset.js
```diff
@@ -8,30 +10,76 @@ class ResetCommand extends BaseCommand {
...
         task: () => (
           new Listr([
             {
-              title: 'Remove Docker containers and associated data',
+              title: 'Remove all services and associated data',
               task: async () => dockerCompose.down(config.toEnvs()),
             },
+            {
+              title: 'Reset config',
+              enabled: () => isHardReset,
```
shumkov, line 72: You need to reset the config only in case if it’s system one.

src/docker/DockerCompose.js
```diff
@@ -200,6 +214,21 @@ class DockerCompose {
...
 
+  /**
+   * Invert list of Docker service names
+   *
+   * @private
+   * @param {Object} envs
+   * @param {string} [invertServiceName]
+   * @return {string[]}
+   */
+  async invertContainerNamesList(envs, invertServiceId = undefined) {
```
shumkov, line 225: Pretty strange function for this class. Do you use it many times in different places?

src/listr/tasks/tenderdash/initTaskFactory.js
```diff
@@ -0,0 +1,61 @@
+const { Listr } = require('listr2');
```
shumkov, line 1: The file name should be equal to a function or a class that it contains.

src/listr/tasks/tenderdash/initTaskFactory.js
```diff
@@ -0,0 +1,61 @@
...
+          const isGenesisPresent = Object.keys(config.get('platform.drive.tenderdash.genesis')).length !== 0;
+
+          const { Volumes: existingVolumes } = await docker.listVolumes();
+          const { COMPOSE_PROJECT_NAME: composeProjectName } = config.toEnvs();
+          const isDataVolumePresent = existingVolumes.find((v) => v.Name === `${composeProjectName}_drive_tenderdash`);
+
+          if (isValidatorKeyPresent && isNodeKeyPresent
+            && isGenesisPresent && isDataVolumePresent) {
+            task.skip('Node already initialized');
+          } else {
```
shumkov, line 34: You can just `return` after `skip` so you don't need `else`

src/commands/reset.js
strophy: Moved this logic to the status command, is that where it should be you think?

src/commands/reset.js
strophy: `down -v` will remove all containers and named/anonymous volumes created by `up` so that won't work if we are trying to remove `--platform-only` containers. The only way to remove some of the containers is with `rm <name>` but `rm -v` will [only remove anonymous volumes](https://docs.docker.com/compose/reference/rm/), it does not remove named volumes. So these are the options:

- Use `down -v` and do not offer `reset --platform-only` option
- Use `stop` then `rm -v` and change volumes to anonymous
- Use `stop` then `rm` and do `docker volume prune` afterwards to remove named volumes

src/docker/DockerCompose.js
strophy: Moved logic to status command

src/listr/tasks/tenderdash/initTaskFactory.js
strophy: Right, so:

- `platform/initTaskFactory.js` contains `platformInitTaskFactory`
- `tenderdash/initTaskFactory.js` contains `tenderdashInitTaskFactory`

Should I rename the functions or the files? Or move everything to another directory called `init` or something?

src/listr/tasks/tenderdash/initTaskFactory.js
strophy: OK, I changed it. But I don't understand what is being returned?

shumkov: Nothing. Just `void`.

src/listr/tasks/tenderdash/initTaskFactory.js
shumkov: File names atm are not equal to functions which they consist. It should be `tenderdashInitTask`

src/commands/reset.js
shumkov: I get it, but I thought you call it in all cases, not just for `--platform-only`

src/commands/reset.js
shumkov: Yep

src/commands/reset.js
strophy: I don't understand what you mean. If we call `down -v` in all cases then we need to sync core again, even if we `reset --platform-only`. But the whole point of `--platform-only` is to stop having to wait for core sync when resetting platform. What am I missing here?

src/commands/reset.js
strophy: Added a new function to check if current config is system config and did some refactoring

src/listr/tasks/tenderdash/initTaskFactory.js
strophy: Why is returning `void` better than if/else? We have already evaluated the condition, so it doesn't cost any processing time and the syntax is a lot clearer (to me)

src/commands/reset.js
shumkov: Not sure that using `pruneVolumes` is a good idea. It will prune all volumes even not related to this docker compose project. What if we just remove volumes using a list of names. I guess it will be safer. Also, if your previous tasks named "Remove platform services and associated data" then I would move this logic there and have just one task

src/commands/reset.js
```diff
@@ -8,37 +10,90 @@ class ResetCommand extends BaseCommand {
...
+      {
+        title: 'Remove all services and associated data',
+        enabled: () => !isPlatformOnlyReset,
+        task: async () => dockerCompose.down(config.toEnvs()),
+      },
+      {
+        title: `Reset config ${config.getName()}`,
+        enabled: () => isHardReset,
+        task: async (ctx, task) => {
+          if (checkSystemConfig(config.getName())) {
```
shumkov, line 75: We usually call such functions like `isSystemConfig`.

src/commands/reset.js
```diff
@@ -8,37 +10,90 @@ class ResetCommand extends BaseCommand {
...
+            resetSystemConfig(configCollection, config.getName(), isPlatformOnlyReset);
+          } else {
+            // eslint-disable-next-line no-param-reassign
+            task.skip('config reset failed, only system configs can be reset');
```
shumkov, line 79: Hm.. It's not actually failed. If we want to fail it we need to call `task.error` or throw an exception. Maybe we just need to check is `--hard` present and config is not system then throw error even before running listr?

strophy: I've changed the wording to "skipped" instead of "failed". But I guess the decision we need to make is if using `--hard` with a non-system config should just skip the config reset task (but still remove the containers and volumes), or if it should fail with an error and not do any tasks. My thinking is it is just part of the command that isn't possible, so skipping and emitting a message with `task.skip()` makes more sense than failing the whole command.

src/commands/reset.js
strophy: Agree that `pruneVolumes` has unintended side effects. But I think collecting a list of volumes and deleting them one by one is an anti-pattern, instead we should be using anonymous volumes for state data and removing with `rm -v`, since all the data we care about (settings and logs) is available through bind mounts.

shumkov: I’m not sure that anonymous volumes are a good idea. It’s easy to lose them but named volumes are a more persistent thing which is matters in the case of node data. Also, I see cases in the future where we are mounting these volumes to different containers.

Why do you think it’s an antipattern. We already have this list in compose config. Do you think the duplication of that list here is a big problem?

src/listr/tasks/tenderdash/initTaskFactory.js
shumkov: It's just a best practice for readability.

https://medium.com/better-programming/why-you-need-to-stop-using-else-statements-5b1fd09dea9e
https://blog.timoxley.com/post/47041269194/avoid-else-return-early

src/commands/reset.js
shumkov: Yeah, agree. But you consider it from the perspective of options validation. You logically can't do `--hard` with `--config=<not system>` so you preventing the whole thing because the input is not valid.

src/listr/tasks/tenderdash/initTaskFactory.js
strophy: Thanks! Reading this helped my understanding a lot :+1:

src/commands/reset.js
strophy: Yeah, I changed my mind now. I think the behaviour you describe is more consistent with other commands, so I've changed the implementation now :+1:

src/commands/reset.js
strophy: OK, fair enough. I was only looking at our current use patterns, which always remove the containers and their data simultaneously, which is what anonymous volumes were designed for. Lucky we have you around to plan for the future ;)

I've refactored it to remove the volumes named in the compose file and cleaned up the listr syntax. We still need to use `dockerode` because there is no way to remove named volumes using docker-compose.

shumkov (approved):
Good job! 👍

src/commands/reset.js
shumkov: Hehe... I got your sarcasm :). Now it looks perfect! We even don't need to maintain the whole list of volumes.

PR #268, opened 2021-02-08
chore: remove evonet-specific code

strophy:
Evonet has been destroyed, so configs and code can be removed from mn-bootstrap.

Evonet no longer exists and should not be offered as an option in mn-bootstrap.

## What was done?
Removed all vars and code specific to evonet. Should we create an issue to allow users to specify another devnet name during setup?

## How Has This Been Tested?
Successfully started testnet node and ran status commands, no errors

## Checklist:


templates/dashd.conf.template
```diff
@@ -39,16 +39,7 @@ zmqpubrawtxlocksig=tcp://0.0.0.0:29998
 [test]
 {{?? it.network === 'local'}}
 regtest=1
-[regtest]{{?? it.network === 'evonet'}}
```
shumkov, line 42: Let's keep it for network `devnet` in case if users decide to create a node for their network.

shumkov (changes requested):

templates/dashd.conf.template
strophy: OK, kept it in there and allowed `devnet` network types again. Do you want to add a system config for devnets as well? I think since it won't work with setup, we can just expect the user to configure a devnet manually.

shumkov: Yeah, exactly!

shumkov (approved):
Looks good! 👍

PR #272, opened 2021-02-13
fix: reset function doesn't wait for volume to be removed

strophy:
Change async function to wait for promises to resolve before moving on to sync code.

Tenderdash was crashing after `mn reset --platform-only` because the Tenderdash volume was deleted, but not recreated. This was a sporadic problem, when the system was under load it did not occur.

## What was done?
Investigation showed the async function to get the docker container and remove it wasn't waiting for the promise to resolve before continuing. I think it is [this problem](https://stackoverflow.com/questions/37576685/using-async-await-with-a-foreach-loop). It works now, but I don't know why. It is almost certainly possible to improve this further using `Promise.all` or using proper chaining syntax.

## How Has This Been Tested?
Test multiple times with low system load, cannot reproduce the bug again.

## Checklist:

### DIFF
--- a/src/commands/reset.js
+++ b/src/commands/reset.js
@@ -71,9 +71,11 @@ class ResetCommand extends BaseCommand {
             .filter((volumeName) => !coreVolumeNames.includes(volumeName))
             .map((volumeName) => `${composeProjectName}_${volumeName}`);
 
+          const volumeRemovePromises = [];
           for (const volumeName of volumeNames) {
-            await (docker.getVolume(volumeName)).remove();
+            volumeRemovePromises.push(docker.getVolume(volumeName).remove());
           }
+          await Promise.all(volumeRemovePromises);
         },
       },
       {
--- a/src/commands/reset.js
+++ b/src/commands/reset.js
@@ -67,10 +67,13 @@ class ResetCommand extends BaseCommand {
 
           const projectvolumeNames = await dockerCompose.getVolumeNames(config.toEnvs());
 
-          await projectvolumeNames
+          const volumeNames = projectvolumeNames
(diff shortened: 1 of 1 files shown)


src/commands/reset.js
```diff
@@ -67,10 +67,13 @@ class ResetCommand extends BaseCommand {
...
-          await projectvolumeNames
+          const volumeNames = projectvolumeNames
             .filter((volumeName) => !coreVolumeNames.includes(volumeName))
-            .map((volumeName) => `${composeProjectName}_${volumeName}`)
-            .forEach(async (volumeName) => docker.getVolume(volumeName).remove());
+            .map((volumeName) => `${composeProjectName}_${volumeName}`);
+
+          for (const volumeName of volumeNames) {
+            await (docker.getVolume(volumeName)).remove();
+          }
```
shumkov, line 76: You can do in map:
```js
const volumeRemovePromises = 
const volume  = docker.getVolume(`${composeProjectName}_${volumeName}`);
return volume.remove();
```

and resulting array of promises resolve like this:

```js
await Promise.all(volumeRemovePromises);
```

strophy: I don't understand the syntax you are using, where is the map in there? Is this a function? Where does the returned object go? I was able to get it working using `Promise.all()` now but I'm not sure this is the result you were expecting.

shumkov: You already iterating through volume names and you can chain one more `map` there which will call `remove` and return promise, so the result variable will be an array of promises that you can put to `Promise.all`

strophy: Fixed, I think. Can't believe this took me so long to understand.

shumkov (approved):
👍

PR #345, opened 2021-05-05
docs: add update docs

strophy:

Upgrade docs were necessary in the case of platform resets. More detailed step-by-step instructions will be [added to the user docs](https://docs.dash.org/en/stable/masternodes/setup-testnet.html#masternode-installation) instead.

## What was done?

- Added steps to update node

## How Has This Been Tested?

tested on testnet

## Breaking Changes

none

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,17 @@ $ mn update
 $ mn start
 ```
 
-If the platform layer has been wiped, you must additionally run `mn reset --platform-only --hard` followed by `mn setup -k <bls-key>` with your existing BLS private key after the first `mn stop` command.
+If the platform layer has been wiped, you must additionally reset platform data:
+
+```bash
+$ mn stop
+$ git checkout master
+$ git pull
+$ mn reset --platform-only --hard
+$ mn update
+$ mn setup -k <bls-key>
+$ mn start
+```
 
 ## Usage
 
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ $ mn update
 $ mn start
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -44,6 +44,18 @@ $ npm install # optional: install CLI dependencies
 $ sudo npm link # optional: link CLI for system-wide execution
 ```
 
+## Update
```
shumkov, line 47: Let's add to TOC?

README.md
```diff
@@ -44,6 +44,18 @@ $ npm install # optional: install CLI dependencies
...
+
+```bash
+$ mn stop
+$ git checkout master
+$ git pull
+$ mn update
+$ mn start
+```
+
+If the platform layer has been wiped, you must additionally run `$ mn reset --platform-only --hard` followed by `mn setup -k <bls-key>` with your existing BLS private key after the first `mn stop` command.
```
shumkov, line 57: Will the setup work if we reset the platform only?

strophy: Actually I'm not sure if `mn setup` is necessary again. Is it possible for the node to use the same `validatorKey` and `nodeKey` on the new network? Would something like this work instead?

```bash
$ mn stop
$ git checkout master
$ git pull
$ mn reset --platform-only
$ mn update
$ mn start
```

I think the other config values (`genesis`, `validators`, `contracts`) would be overwritten by the new values from the `git pull` command? Or would we need to write migrations for this?

shumkov (approved):
👍

PR #362, opened 2021-05-21
feat!: enable and request debug output for setup local

strophy:

Enable debug output flag for a network created with `setup local`

## What was done?

- Add boolean `debug` flag
- Add step to query debug if not specified
- Set core and platform debug appropriately in config

## How Has This Been Tested?

Tested local and in CI

## Breaking Changes

`setup local` command requests debug setting

## Checklist:

### DIFF
--- a/src/commands/setup.js
+++ b/src/commands/setup.js
@@ -33,7 +33,7 @@ class SetupCommand extends BaseCommand {
       'operator-bls-private-key': operatorBlsPrivateKey,
       'funding-private-key': fundingPrivateKeyString,
       'node-count': nodeCount,
-      debug,
+      'debug-logs': debugLogs,
       verbose: isVerbose,
     },
     generateBlsKeys,
@@ -96,7 +96,7 @@ class SetupCommand extends BaseCommand {
         preset,
         nodeType,
         nodeCount,
-        debug,
+        debugLogs,
         externalIp,
         operatorBlsPrivateKey,
         fundingPrivateKeyString,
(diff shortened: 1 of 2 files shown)


src/commands/setup.js
```diff
@@ -129,6 +131,7 @@ SetupCommand.flags = {
   'funding-private-key': flagTypes.string({ char: 'p', description: `private key with more than ${MASTERNODE_DASH_AMOUNT} dash for funding collateral` }),
   'node-count': flagTypes.integer({ description: 'number of nodes to setup', default: null }),
   verbose: flagTypes.boolean({ char: 'v', description: 'use verbose mode for output', default: false }),
+  debug: flagTypes.boolean({ char: 'd', description: 'enable debug output', default: null }),
```
shumkov, line 134: output -> logs

shumkov, line 134: debug-logs

shumkov (changes requested):

shumkov (dismissed):
👍

src/listr/tasks/setup/setupLocalPresetTaskFactory.js
```diff
@@ -41,6 +41,19 @@ function setupLocalPresetTaskFactory(
           });
         },
       },
+      {
+        title: 'Enable debug logs',
+        enabled: (ctx) => ctx.debugLogs === null,
```
shumkov, line 46: This won't work because it will be false by default. I guess you need to set `null` as default.

strophy: Ah, I see how it works now! So it can basically have three meaningful states now: `true`, `false` and `null`?

shumkov (approved):
👍

shumkov (approved):
👍

PR #368, opened 2021-05-27
feat: add Tor service

strophy:

Core devs have requested a Tor service option. This PR is based on [this forum post](https://www.dash.org/forum/threads/how-to-tor.50363/) by xkcd.

## What was done?

- Added Tor service as override service
- Added config to enable and parametrize Tor service

## How Has This Been Tested?

Testing on testnet. Following concerns:
- [x] `[warn] You specified a public address '0.0.0.0:9040' for TransPort. Other people on the Internet might find your computer and use it as an open proxy. Please don't allow this unless you have a good reason.`
  How do we want this configured?
- [x] `[warn] You specified a public address '0.0.0.0:9050' for SocksPort. Other people on the Internet might find your computer and use it as an open proxy. Please don't allow this unless you have a good reason.`
  How do we want this configured?
- [x] `[warn] You have a ControlPort set to accept connections from a non-local address.  This means that programs not running on your computer can reconfigure your Tor.  That's pretty bad, since the controller protocol isn't encrypted!  Maybe you should just listen on 127.0.0.1 and use a tool like stunnel or ssh to encrypt remote connections to your control port.`
  I believe this is secure because only containers in the Docker network can access this port. Let me know if I should try to implement stunnel or ssh anyway.
- [x] Tor service starts conditionally based on updated logic in this PR. Test image `strophy/dashd:tor` is currently specified in [this PR](https://github.com/dashpay/dash/pull/4182).
- [ ] How can I test if the hidden service is working?

## Breaking Changes

- Tor is enabled by default on testnet

## Checklist:

- [ ] I have performed a self-review of my own code


configs/system/base.js
```diff
@@ -38,6 +38,19 @@ module.exports = {
...
+      enable: true,
+      docker: {
+        image: 'dperson/torproxy',
+      },
+      proxy: {
+        port: 9050,
+      },
+      control: {
+        port: 9051,
+        password: 'torcontrolpassword',
```
shumkov, line 51: Not sure it's a good idea to have a default value. Maybe generate it on setup?

docker-compose.tor.yml
```diff
@@ -0,0 +1,21 @@
+version: '3.7'
+
+services:
+  tor:
```
shumkov, line 4: core_tor

configs/system/base.js
```diff
@@ -38,6 +38,19 @@ module.exports = {
         image: 'dashpay/sentinel:1.5.0',
       },
     },
+    tor: {
+      enable: true,
```
shumkov, line 42: Are you sure we want to have it enabled by default for all presets?

configs/system/base.js
strophy: We use a default value for dashd as well. Let's create a separate issue for randomizing passwords on startup?

configs/system/base.js
strophy: No, I'm not sure. I am waiting for guidance from @PastaPastaPasta about when this service should be enabled

configs/system/base.js
shumkov: Sounds good!

configs/system/base.js
shumkov: Ok. I pretty sure we don't need it for the local network.

PastaPastaPasta:
See https://github.com/dperson/torproxy/issues/44 about some of the questions you have when it comes to the warnings

configs/system/base.js
shumkov, line 42: I would keep it disabled for default and enable it in specific configs (testnet atm)

strophy:
I am stuck unable to connect to the hidden service. The `proxy` and `torcontrol` config settings seem to be working, we can connect to Tor properly and I see the following logs from the dash `core` container:
```
2021-06-10T11:14:51Z torcontrol thread start
2021-06-10T11:14:51Z tor: Error connecting to Tor control socket
2021-06-10T11:14:51Z tor: Not connected to Tor control port 172.19.0.2:9051, trying to reconnect
2021-06-10T11:14:52Z tor: Successfully connected!
2021-06-10T11:14:52Z tor: Connected to Tor version 0.4.3.5
2021-06-10T11:14:52Z tor: Supported authentication method: HASHEDPASSWORD
2021-06-10T11:14:52Z tor: Using HASHEDPASSWORD authentication
2021-06-10T11:14:52Z tor: Authentication successful
2021-06-10T11:14:52Z tor: ADD_ONION successful
2021-06-10T11:14:52Z tor: Got service ID zqhsuwy372ez4m2a, advertising service zqhsuwy372ez4m2a.onion:19999
2021-06-10T11:14:52Z tor: Cached service private key to /dash/.dashcore/testnet3/onion_private_key
```

But configuring a Dash Core wallet to connect to this service with following `dash.conf`:
```
testnet=1
onlynet=onion
proxy=127.0.0.1:9150
[test]
addnode=zqhsuwy372ez4m2a.onion:19999
```
Results in:
```
2021-06-10T12:05:49Z Socks5() connect to zqhsuwy372ez4m2a.onion:19999 failed: general failure
```
I think this is a problem in Tor config, because when the dashmate node is down, the error changes to:
```
2021-06-10T11:25:37Z Socks5() connect to zqhsuwy372ez4m2a.onion:19999 failed: host unreachable
```
This could be a problem for the Tor service to reach the `dashd` service? Most configuration examples I found always run both services on the same host, should I try this? Or can someone help me debug Tor in a container?

UdjinM6:
Make sure to specify the right port
```
proxy=127.0.0.1:9050  # use tor service proxy
```
vs
```
proxy=127.0.0.1:9150  # use tor browser proxy
```

kxcd:
Both ports are right, but in the context of MN server, 9050 is the right and tor the daemon runs on 9050.

strophy:
Thanks guys, finally had a good look at this again today. The proxy discussion above was beside the point, I was using the browser proxy under Windows, but unable to connect.

Since picking this up again, I first tested dashd + Tor through a manual install without use of Docker or dashmate. This worked easily enough, and Tor was successfully configured by dashd through the ControlPort, and I was able to use the v2 hidden service address shown in the dashd logs to connect and sync from a remote node.

I then implemented the same configuration in dashmate and saw the same hidden service output in the dashd logs, but a remote node was unable to connect in this configuration. I'm not sure how to debug this, we would need someone familiar with debugging the tor control specifications and what ports are needed. Regardless, I think the current implementation of the control specifications is not workable because it sets up a Tor v2 address. These addresses will [stop working](https://blog.torproject.org/v2-deprecation-timeline) on or soon after 2021-10-15. Backporting [Bitcoin #19954](https://github.com/bitcoin/bitcoin/pull/19954) will allow us to attempt setting up control from dashd again.

Finally, I modified the Tor container configuration to set up a hidden service manually. This works fine by simply running `dashmate config:set core.tor.enable true`. The hidden service will be set up, and the address appears in `/var/lib/tor/hidden_service/hostname` in the Tor container. It is possible to sync a remote node against this address, even with `onlynet=onion` enabled. dashd has no knowledge that it is being accessed over Tor in this configuration.

PastaPastaPasta:
> These addresses will [stop working](https://blog.torproject.org/v2-deprecation-timeline) on or soon after 2021-10-15. Backporting [Bitcoin #19954](https://github.com/bitcoin/bitcoin/pull/19954) will allow us to attempt setting up control from dashd again.

We backported this in https://github.com/dashpay/dash/pull/4181

PR #379, opened 2021-06-11
feat: support minor Core updates

strophy:

`dashmate update` should be able to download minor Core updates

## What was done?

- Set less precise Core image `0.17`
- Update sentinel to 1.6.0
- Update migrations
- Fix error in 0.19.1 migration

## How Has This Been Tested?

Tested migration from master to this branch by incrementing package version

## Breaking Changes

None

## Checklist:

### DIFF
--- a/configs/system/base.js
+++ b/configs/system/base.js
@@ -7,7 +7,7 @@ module.exports = {
   group: null,
   core: {
     docker: {
-      image: 'dashpay/dashd:0.17.0.0-rc5',
+      image: 'dashpay/dashd:0.17.0.3',
     },
     p2p: {
       port: 20001,
@@ -35,7 +35,7 @@ module.exports = {
     },
     sentinel: {
       docker: {
-        image: 'dashpay/sentinel:1.5.0',
+        image: 'dashpay/sentinel:1.6.0',
       },
     },
     devnetName: null,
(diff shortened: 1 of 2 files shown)


configs/system/base.js
```diff
@@ -7,7 +7,7 @@ module.exports = {
   group: null,
   core: {
     docker: {
-      image: 'dashpay/dashd:0.17.0.0-rc5',
+      image: 'dashpay/dashd:0.17.0.3',
```
shumkov, line 10: I would recommend just setting the `0.17` tag so users can do `dashmate update` instead of releasing dashmate for every minor Core release.

configs/system/base.js
```diff
@@ -35,7 +35,7 @@ module.exports = {
     },
     sentinel: {
       docker: {
-        image: 'dashpay/sentinel:1.5.0',
+        image: 'dashpay/sentinel:1.6.0',
```
shumkov, line 38: The same for sentinel.

shumkov (changes requested):

configs/system/base.js
shumkov: We don't have less precise tags for sentinel?

shumkov:
@strophy I guess we don't care to much about Sentinel. Let's keep it as it is for now.

shumkov (approved):
👍

PR #388, opened 2021-06-23
feat: enable mainnet

strophy:

Masternode owners have expressed interest in testing dashmate on mainnet.

## What was done?

- Add mainnet config with platform disabled
- Disable tenderdash init on mainnet

## How Has This Been Tested?

- Tested locally
- Ongoing testnet on VPS

## Breaking Changes

None

## Checklist:


templates/core/dash.conf.dot
```diff
@@ -26,11 +26,12 @@ externalip={{=it.externalIp}}
 whitelist={{=it.externalIp}}
 {{?}}
 
-# Indices
+{{? it.network !== 'mainnet'}}# Indices
```
thephez, line 29: I thought indices were required on mainnet. Or is this because they're enforced automatically?

strophy: I think the necessary indices will be enforced automatically. I have tried syncing this branch on a 55GB Vultr VPS with following results:

- Indices enabled: drive full at block 1303209
- Indices disabled: sync complete using 37GB

thephez: I think txindex is the only one currently required by masternodes (attempting to confirm this). If that's accurate and platform requires additional indices, we'll probably need to leave them in here. 

Also, if that is the case, I suppose the next release of Core should add them to the list of enforced ones for masternodes. @PastaPastaPasta @UdjinM6

strophy: Yes, I agree. This branch is currently for some enthusiastic users to try setting up mainnet nodes with dashmate on their existing hardware (typically 55GB Vultr nodes). This condition check for mainnet can be removed once platform is supported on mainnet and these indices are required again.

kxcd: Yes, when you configure the blskey it automatically enables several other dependent options I believe the additional index is one of them and no doubt once Evo is launched to mainnet those additional indices will be added to the default list.

OT: It makes for an interesting question on upgrade from existing network to Evo, I suppose each exiting legacy mnode will have to reindex (expensive) their chains.

antouhou (dismissed):

antouhou (approved):

PR #428, opened 2021-09-17
fix: drive log files were missing upon setup

jawid-h:

In some situations Drive log files were missing upon setup of the nodes, this leads to Docker infinitely restarts container.

## What was done?

- implemented `touchFile` function
- touching or creating an empty log file on `setup`

## Breaking Changes

None

## Checklist:


shumkov (changes requested):
This logic already exists in `startNodeTask` and I think it’s better place since the problem won’t be solved if you don’t use `setup local`. I guess, the thing is that this logic is missed in `startGroupNodes` which is used in case of local network.

src/listr/tasks/setup/setupLocalPresetTaskFactory.js
```diff
@@ -139,8 +139,11 @@ function setupLocalPresetTaskFactory(
                     });
                   }
 
-                  config.set('platform.drive.abci.log.prettyFile.path', path.join(os.tmpdir(), `/drive_pretty_${nodeIndex}.log`));
-                  config.set('platform.drive.abci.log.jsonFile.path', path.join(os.tmpdir(), `/drive_json_${nodeIndex}.log`));
+                  const drivePrettyLogFile = path.join(os.tmpdir(), `/drive_pretty_${nodeIndex}.log`);
+                  const driveJsonLogFile = path.join(os.tmpdir(), `/drive_json_${nodeIndex}.log`);
```
shumkov: Remove slash

shumkov (approved):
I hope it works :)

PR #437, opened 2021-09-29
fix: issues related to keeping log files in /tmp

strophy:

Log files were hard coded to use `os.tmpdir()`, i.e. `/tmp` on Linux or `C:\\Users\\<username>\\AppData\\Local\\Temp\\` on Windows. This caused problems under Linux, because the `/tmp` dir is cleaned once daily and/or on restart. When dashmate was started again, Docker would create the non-existent file with `root` owner, making it unwritable in the container and impossible for an unprivileged user to manage.

## What was done?

Moved log files to `os.homedir()/.dashmate/<config>/logs`.

## How Has This Been Tested?

Tested under Windows and Linux

## Checklist:


strophy:
I'm not sure this is the correct approach, because we also specify the `homeDir` in the DI container [here](https://github.com/dashevo/dashmate/blob/move-log-files/src/createDIContainer.js#L81). But because the system configs are mostly static I couldn't figure out a way to access the DI container scope when processing the system configs. So I just made a [constant ](https://github.com/dashevo/dashmate/blob/move-log-files/src/constants.js#L28)instead, which means the logs will be stored here even if the user specifies a different datadir. Is there a better way?

src/constants.js
```diff
@@ -22,6 +25,8 @@ const quorumTypes = {
 
 const MASTERNODE_DASH_AMOUNT = 1000;
 
+const HOME_DIR_PATH = path.resolve(os.homedir(), '.dashmate');
```
shumkov: We have it already defined in the container https://github.com/dashevo/dashmate/blob/e468f94d1b9458ea5d1c429f0baa8073316ac419/src/createDIContainer.js#L81

We could probably define it here and use this constant in the container?

strophy: OK, I have changed the DI container so it now gets the variable as a constant. Does that look right?

configs/system/base.js
```diff
@@ -91,11 +91,11 @@ module.exports = {
           },
           prettyFile: {
             level: 'silent',
-            path: path.join(os.tmpdir(), '/base-drive-pretty.log'),
+            path: path.join(HOME_DIR_PATH, '/base/logs/', '/drive-pretty.log'),
```
shumkov: The point of using `join` is to do not use the separator (`/`) which is different on different OS. So you need to pass all segments of the path as a separate arguments of `join` function. Like `path.join(HOME_DIR_PATH, 'base', 'logs', 'drive-pretty.log')`

src/constants.js
shumkov: So we don't allow to configure the dashmate home dir with `DASHMATE_HOME_DIR` env anymore?

src/createDIContainer.js
```diff
@@ -70,16 +71,14 @@ const enableCoreQuorumsTaskFactory = require('./listr/tasks/setup/local/enableCo
...
     injectionMode: InjectionMode.CLASSIC,
   });
 
   /**
    * Config
    */
-  const homeDirPath = options.DASHMATE_HOME_DIR ? options.DASHMATE_HOME_DIR : path.resolve(os.homedir(), '.dashmate');
-
   container.register({
     homeDirPath: asValue(homeDirPath),
```
shumkov: It's probably better to remove this value from the container and use constants where it was used.

shumkov (changes requested):

src/constants.js
strophy: Right, this option is no longer available.

shuplenkov (dismissed):

shuplenkov (dismissed):

src/commands/config/envs.js
```diff
@@ -3,14 +3,15 @@ const path = require('path');
 
 const { flags: flagTypes } = require('@oclif/command');
 
+const { HOME_DIR_PATH: homeDirPath } = require('../../constants');
```
shumkov: I would use the constant name to keep more visible when you read the code.

src/constants.js
shumkov: Why just don’t put this logic here?

shumkov (changes requested):

src/commands/config/envs.js
strophy: OK, I changed all of the destructuring to just use the constant instead, does [this](https://github.com/dashevo/dashmate/pull/437/commits/e8b83f81283e6966685d52c5a9464ec39578898a) look right? Or should it only be in `envs.js`?

src/constants.js
strophy: Because if the value is potentially variable, then it is by definition not a constant anymore and doesn't belong in `constants.js`

strophy:
Actually I just realised that nothing is written to these log files anyway, because we only modify `platform.drive.abci.log.stdout.level` [here](https://github.com/dashevo/dashmate/blob/move-log-files/src/listr/tasks/setup/setupLocalPresetTaskFactory.js#L138). Shouldn't we be setting the `platform.drive.abci.log.prettyFile.level` and `platform.drive.abci.log.jsonFile.level` instead? These are currently set to `silent` even if the user selects enable logging in `dashmate setup`...

src/config/configFile/ConfigFileJsonRepository.js
```diff
@@ -5,6 +5,8 @@ const Ajv = require('ajv');
 const Config = require('../Config');
 const ConfigFile = require('./ConfigFile');
 
+const { CONFIG_FILE_PATH: configFilePath } = require('../../constants');
```
shumkov: Let's use original constant name

src/docker/DockerCompose.js
```diff
@@ -13,18 +13,19 @@ const ServiceAlreadyRunningError = require('./errors/ServiceAlreadyRunningError'
...
    * @param {Docker} docker
    * @param {StartedContainers} startedContainers
-   * @param {string} homeDirPath
    */
-  constructor(docker, startedContainers, homeDirPath) {
+  constructor(docker, startedContainers) {
     this.docker = docker;
     this.startedContainers = startedContainers;
-    this.homeDirPath = homeDirPath;
+    this.homeDirPath = HOME_DIR_PATH;
```
shumkov: Having constant in upper sope you don't need this property anymore

src/config/configFile/ConfigFileJsonRepository.js
```diff
@@ -14,11 +16,10 @@ const packageJson = require('../../../package.json');
 
 class ConfigFileJsonRepository {
   /**
-   * @param {string} configFilePath
    * @param {migrateConfigFile} migrateConfigFile
    */
-  constructor(configFilePath, migrateConfigFile) {
-    this.configFilePath = configFilePath;
+  constructor(migrateConfigFile) {
+    this.configFilePath = CONFIG_FILE_PATH;
```
shumkov: There is no point to have this property. You can use constant directly

shumkov (approved):
LGTM! 👍

Issue #430, opened 2021-09-20
dashmate hangs during restart


dashameter:
when running `dashmate group:stop` and later `dashmate group:start` dashmate hangs indefinitely at `forcing nodes to sync`

```
❯ Start local nodes
  ❯ Starting nodes
    ❯ Starting local_1 node
      ✔ Check node is not started
      ✔ Start services
      ✖ Force nodes to sync
    ❯ Starting local_2 node
      ✔ Check node is not started
      ✔ Start services
      ✖ Force nodes to sync
    ❯ Starting local_3 node
      ✔ Check node is not started
      ✔ Start services
      ✖ Force nodes to sync
    ❯ Starting local_seed node
      ✔ Check node is not started
      ✔ Start services
      ✖ Force nodes to sync
  ◼ Wait for Core peers to be connected
  ◼ Adjust Core mock time
  ◼ Start a miner
```

dash-maverick:
Fix for that issue is currently under progress

shumkov:
Fixed in #428 and #431

SamKirby22:
@dashameter, as per discord convo, please retest. Should be fixed due to:

https://github.com/dashevo/dashmate/pull/437 (confirmed by Strophy)
+
https://github.com/dashevo/dashmate/pull/428
https://github.com/dashevo/dashmate/pull/431 (as above confirmed by shumkov)

dashameter:
passed retest :+1:

PR #448, opened 2021-10-06
fix: block database contains a block from the future

strophy:

Dashmate local networks were failing after running for a longer period of time and/or with very short block times (e.g. 10sec on a 4th gen. i7, or 20sec in CI). #430 is also possibly related, and restarting the host without stopping dashmate first may also trigger the bug according to some reports. Core logs generally showed something similar to this:
```
2021-09-29T14:04:08.498538400Z 2021-09-29T14:04:08Z : The block database contains a block which appears to be from the future. This may be due to your computer's date and time being set incorrectly. Only rebuild the block database if you are sure that your computer's date and time are correct.
2021-09-29T14:04:08.498542600Z Please restart with -reindex or -reindex-chainstate to recover.
2021-09-29T14:04:08.498615500Z 2021-09-29T14:04:08Z Aborted block database rebuild. Exiting.
```

## What was done?

The mocktime feature would simulate 2.5 minute blocks to ensure quorum requests still worked normally. When stopping and then restarting the network, the mocktime would not be set when scanning the block database, causing an error. This PR stores the blocktime to config on shutdown for local networks, and applies the mocktime (if present) to the config file on startup.

## How Has This Been Tested?

- Tested with local network
- Tested to ensure testnet config stop/start still works
- Needs testing in CI still

## Breaking Changes

None

## Checklist:

### DIFF
--- a/src/listr/tasks/stopNodeTaskFactory.js
+++ b/src/listr/tasks/stopNodeTaskFactory.js
@@ -2,10 +2,13 @@ const { Listr } = require('listr2');
 
 /**
  * @param {DockerCompose} dockerCompose
- *
+ * @param {createRpcClient} createRpcClient
  * @return {stopNodeTask}
  */
-function stopNodeTaskFactory(dockerCompose) {
+function stopNodeTaskFactory(
+  dockerCompose,
+  createRpcClient,
+) {
   /**
    * Stop node
    * @typedef stopNodeTask
@@ -13,10 +16,7 @@ function stopNodeTaskFactory(dockerCompose) {
    *
    * @return {Listr}
    */
-  function stopNodeTask(
-    config,
-    createRpcClient,
-  ) {
+  function stopNodeTask(config) {
     return new Listr([
       {
         title: 'Save mock time',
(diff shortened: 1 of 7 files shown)


src/listr/tasks/stopNodeTaskFactory.js
```diff
@@ -13,8 +13,26 @@ function stopNodeTaskFactory(dockerCompose) {
    *
    * @return {Listr}
```
shumkov: You should add createRpcClient to the factory argument.

shumkov (changes requested):

src/listr/tasks/stopNodeTaskFactory.js
```diff
@@ -15,6 +18,21 @@ function stopNodeTaskFactory(dockerCompose) {
    */
   function stopNodeTask(config) {
     return new Listr([
+      {
+        title: 'Save mock time',
```
shumkov: Save Core node time

shumkov (commented):
Could you please also rename the task “Adjust mock time” to something like “Mock Core node time”

shumkov (approved):
Very elegant solution! Good job! 👍

PR #466, opened 2021-11-17
fix: wallet access on seed node in local test network

Arne96R:
Wallet access on seed node (via rpcs) in local network now works. Changed the docker-compose file so it does not feed the `-masternodeblsprivkey` flag to `dashd` when a node is not a masternode. As a result wallet functionality is not disabled in seed node, which it previously was.

Fixes issue [https://github.com/dashevo/dashmate/issues/465](url). 

## What was done?
Change docker-compose command for core containers. Use bash to first check whether node is a masternode, if not a masternode then do not add the `-masternodeblsprivkey`flag to the dashd command.

## How Has This Been Tested?
Used `dashmate setup local`, to set up local network. Then use RPC to check whether wallet functionality on seed node works, which it does. Starting and stopping of nodes works fine.

No known effects on other code.

## Checklist:

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,9 +10,11 @@ services:
     volumes:
       - core_data:/dash
       - ${DASHMATE_HOME_DIR:?err}/${CONFIG_NAME:?err}/core/dash.conf:/dash/.dashcore/dash.conf
-    command:
-      - dashd
-      - -masternodeblsprivkey=${CORE_MASTERNODE_OPERATOR_PRIVATE_KEY}
+    command: bash -c "if [[ $CORE_MASTERNODE_ENABLE = true ]]; then
+                   dashd -masternodeblsprivkey=${CORE_MASTERNODE_OPERATOR_PRIVATE_KEY};
+                else
+                   dashd;
+                fi"
 
 volumes:
   core_data:


shumkov (changes requested):
Best practise is to create an entrypoint with condition logic instead of putting bash scripts right to the cmd. I guess we need to start with dash core entrypoint and read environment variable and add the argument mn key argument when it's present.

Arne96R:
agreed @shumkov , but that would require changing dashcore [entrypoint.sh](https://github.com/dashpay/dash/blob/master/docker/docker-entrypoint.sh) which I wanted to avoid since I thought this is merely a dashmate issue. But I do agree that would be more neat.

Shall I then add an environment section to this docker-compose file, and add a PR on [entrypoint.sh](https://github.com/dashpay/dash/blob/master/docker/docker-entrypoint.sh) to include the logic for checking the existence of that env var? can do that, have it ready.

also, docker hub says [this repo](https://github.com/dashpay/docker-dashd) is the official repo, but in practice [this image](https://github.com/dashpay/dash/tree/master/docker) is used right? at least by dashmate.

shumkov:
@Arne96R 

> Shall I then add an environment section to this docker-compose file, and add a PR on entrypoint.sh to include the logic for checking the existence of that env var? can do that, have it ready.

Yeah, I believe it's the right way.

> also, docker hub says this repo is the official repo, but in practice this image is used right? at least by dashmate.

For me, it's confusing too. @strophy could you please explain where the docker image is building?


One more thing, this repo is deprecated and moved to multi-package repository https://github.com/dashevo/platform/tree/v0.22-dev/packages/dashmate, so please recreate this PR there.

